PR #3, opened 2022-07-20
feat: skeleton code for doracle

### LINKED ISSUE #2: setup initial structure of doracle
setting up initial structure of doracle process.

it may include the followings:
- cmd for daemon
- graceful shutdown of process
- config for doracle

0xHansLee:
I copied some skeleton codes for doracle from the previous oracle, [panacea-coracle](https://github.com/medibloc/panacea-coracle).
(including `Makefile` and some updates of `ci.yaml`)

We don't need the followings anymore: 
- http listening address
- TLS cert
- Authentication

We can make doracle based on this skeleton. If you have any idea(or advice) of refactoring, improved structure, or more fundamental structures, leave comments pls!


0xHansLee:
I convert this PR to draft since I have to fix the lint issue in ci.

config/config.go
```diff
@@ -0,0 +1,54 @@
...
+	OracleMnemonic string `mapstructure:"doracle-mnemonic"`
+}
+
+type PanaceaConfig struct {
+	GRPCAddr string `mapstructure:"grpc-addr"`
+}
+
+type EnclaveConfig struct {
+	Enable                  bool   `mapstructure:"enable"`
+	AttestationProviderAddr string `mapstructure:"attestation-provider-addr"`
```
gyuguen: I'm not sure if we need this setting.
Let's keep it for now and check later.

go.mod
```diff
@@ -1,3 +1,108 @@
 module github.com/medibloc/panacea-doracle
 
 go 1.17
+
+require (
+	github.com/cosmos/cosmos-sdk v0.45.4
```
gyuguen, line 6: Wouldn't we need panacea-core?
Umm, Is it because our fix hasn't been applied to the master branch yet?

store/storage.go
```diff
@@ -0,0 +1,8 @@
+package store
+
+type Storage interface {
```
gyuguen, line 3: I think this interface will be changed for IPFS.

gyuguen (commented):
Thank you.
I have finished this version code.
Additional corrections will be confirmed later.

config/config.go
0xHansLee: Same to me.
I think the `Enable` is needed for running and testing the doracle in SGX-disabled environment.
And `AttestationProviderAddr` will be used for verification of remote report I think.
I'll make this clear when I add SGX related things

go.mod
0xHansLee: The reason why I add cosmos-sdk is for interface registry when setting gRPC connection.
As you said, doracle also needs panacea-core.
This PR is super basic skeleton, so I will add panacea-core in the following PRs.

store/storage.go
0xHansLee: I see. I will consider to make change for more general use.

.github/workflows/ci.yaml
```diff
@@ -15,12 +15,27 @@ jobs:
...
+          go-version: 1.17
         id: go
 
       - name: Checkout code into the Go module directory
         uses: actions/checkout@v2
 
+      - name: Tidy
+        run: |
+          rm -f go.sum
+          go mod tidy -go=1.16 && go mod tidy -go=1.17
```
youngjoon-lee: I don't see any advantage of having this in CI. As devs, we should tidy the go.sum before opening PRs so that CI and outside users don't need to be bothered of it.

Makefile
```diff
@@ -0,0 +1,25 @@
+export GO111MODULE = on
+
+GO ?= go
+
+build_tags := $(strip $(BUILD_TAGS))
+BUILD_FLAGS := -tags "$(build_tags)"
+
+OUT_DIR = ./build
+
+.PHONY: all build test install ego-sign clean
```
youngjoon-lee:
```suggestion
.PHONY: all build test install clean
```

config/toml.go
```diff
@@ -0,0 +1,88 @@
...
+const DefaultConfigTemplate = `# This is a TOML config file.
+# For more information, see https://github.com/toml-lang/toml
+
+###############################################################################
+###                           Base Configuration                            ###
+###############################################################################
+
+log-level = "{{ .BaseConfig.LogLevel }}"
+oracle-mnemonic = "{{ .BaseConfig.OracleMnemonic }}"
+data-encryption-key-file = "{{ .BaseConfig.DataEncryptionKeyFile }}"
```
youngjoon-lee: Please check if this `data-encryption-key-file` is necessary for doracles.

config/toml.go
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+###                         Panacea Configuration                           ###
+###############################################################################
+
+[panacea]
+
+grpc-addr = "{{ .Panacea.GRPCAddr }}"
```
youngjoon-lee, line 31: I guess we need the `rpc-addr` (26657 stuff) instead of this now, especially for subscribing Tendermint events..

config/toml.go
```diff
@@ -0,0 +1,88 @@
...
+
+###############################################################################
+###                         Enclave Configuration                           ###
+###############################################################################
+
+[enclave]
+
+enable = {{ .Enclave.Enable }}
+attestation-provider-addr = "{{ .Enclave.AttestationProviderAddr }}"
```
youngjoon-lee: We don't need this `attestation-provider-addr` anymore for doracles. We are gonna use the Azure DCAP client that is used automatically by `enclave.VerifyReport(...)`. I've already installed all libraries for Azure DCAP client on our Azure VMs.
https://github.com/medibloc/doracle-poc/blob/dd4bacaaa029e7217872279ad87b19d4412f48f0/README.md?plain=1#L20

go.mod
youngjoon-lee: Ja. When you import the panacea-core, please refer to https://medibloc.gitbook.io/panacea-core/guide/installation#import-panacea-core-as-a-go-dependency for specifying `replace` directives (Please double-check package versions in the `replace` directives).

server/server.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/medibloc/panacea-doracle/server/service"
+	"net/http"
+	"os"
+	"os/signal"
+	"time"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func Run(conf *config.Config) error {
```
youngjoon-lee, line 17: As I remember, we don't need this HTTP server anymore. Instead, we are gonna run a Tendermint event subscriber. Because the subscriber will be implement in other PRs, it's enough for you to make this blank.

youngjoon-lee: Well, we will probably need to have a HTTP endpoint or so for some admin purposes, but not sure for now.

config/toml.go
gyuguen: ~~It seems that we need to communicate with grpc to get the proof information of the query.~~

Please ignore the above comment
 I was mistaken😭

.github/workflows/ci.yaml
0xHansLee: I see. I'll remove this step.

config/toml.go
0xHansLee: Ok. I think we need gRPC connection to broadcast Txs for voting stuffs as well.

go.mod
0xHansLee: Thank you for your comments. When I import panacea-core, I'll check it!

gyuguen (approved):
LGTM 👍

config/toml.go
0xHansLee: cosmos sdk said that the preferred way to broadcast Tx is to use gRPC, so let's keep gRPC as well as rpc (for subscription)
ref) https://docs.cosmos.network/main/run-node/txs.html#broadcasting-a-transaction-2

server/server.go
0xHansLee: As you said and we discussed off-line, we decided to keep use HTTP server for some future admin purpose.

.github/workflows/ci.yaml
0xHansLee: It would have been nice if we could detect a situation that is not tidy in advance and cause the github actions to fail. However, since this step is just to run the `go mod tidy`, not to fail, we decided to find another better way.

server/server.go
```diff
@@ -0,0 +1,62 @@
...
+
+	log.Info("terminating HTTP server")
+	ctxTimeout, cancel := context.WithTimeout(context.Background(), time.Second*10)
+	defer cancel()
+
+	if err := server.Shutdown(ctxTimeout); err != nil {
+		return fmt.Errorf("error occurs while server shutting down: %w", err)
+	}
+
+	return nil
```
youngjoon-lee, line 61: Good to have this HTTP server as you said in https://github.com/medibloc/panacea-doracle/pull/3/files#r929439137.
Then, please keep in mind that this code block needs to be extracted to the `main()` function in the future, because we are going to run one more goroutine for subscribing Tendermint events. Then, after getting a OS signal (SIGINT, or so), you need to shutdown the subscriber as well as the HTTP server.

youngjoon-lee, line 61: Good to have this HTTP server as you said in https://github.com/medibloc/panacea-doracle/pull/3/files#r929439137.
Then, please keep in mind that this code block needs to be extracted to the `main()` function in the future, because we are going to run one more goroutine for subscribing Tendermint events. Then, after getting a OS signal (SIGINT, or so), you need to shutdown the subscriber as well as the HTTP server.

youngjoon-lee (approved):
lgtm

server/server.go
0xHansLee: Yes. I remember we made an issue related refactoring graceful shutdown before (https://github.com/medibloc/panacea-coracle/issues/34). I will move this graceful shutdown to `main()` in the following PR, because I think it also needs to be done in code structuring. Thank you for reminding me!

store/storage.go
inchori: I have a question about this code part and comments. In near future, we will use IPFS as a storage. So we are not need to use interface anymore?

0xHansLee: Even if we use IPFS as a storage, I think we can utilize this interface, which would be useful when mocking storage for test.

inchori:
LGTM 👍

PR #4, opened 2022-07-25
feat: add gen-oracle-key command

0xHansLee:
A command to generate a new oracle key has been implemented. 
The generated private key is sealed and stored as `oracle_priv_key.sealed` and the resulting public key and its remote report would be stored under `oracle_pub_key.json`.


cmd/doracled/cmd/root.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	NodePrivKeyFilePath   = "/data/node_priv_key.sealed"
+	OraclePrivKeyFilePath = "/data/.doracle/oracle_priv_key.sealed"
+	OraclePubKeyFilePath  = "/data/.doracle/oracle_pub_key.json"
+)
```
gyuguen: I think these consts can be used in many places, not just cmd.
Well, but I'll think a little more about where I can move it.

config/config.go
```diff
@@ -13,7 +13,8 @@ type Config struct {
 
 type BaseConfig struct {
 	LogLevel       string `mapstructure:"log-level"`
-	OracleMnemonic string `mapstructure:"doracle-mnemonic"`
+	OracleMnemonic string `mapstructure:"oracle-mnemonic"`
+	ListenAddr     string `mapstructure:"listen_addr"`
```
gyuguen, line 13: It seems doracle is not accepting api requests, Where do I use this setting?

gyuguen (commented):
Thank you.
i left a few comment

cmd/doracled/cmd/root.go
0xHansLee: Agreed. How about to make `types` package and move this consts in `consts.go`? or `paths.go`?

config/config.go
0xHansLee: When the listening address is not given to HTTP server, the server has default listening addr, ":http" (port 80). ([ref](https://pkg.go.dev/net/http#Server))
So I add listening addr to config, and it can be used for future admin purpose (https://github.com/medibloc/panacea-doracle/pull/3#discussion_r929439137).

gyuguen: OK. I got it.

cmd/doracled/cmd/root.go
gyuguen: Yes. That will be fine. Any name seems fine, I like `consts.go`.

youngjoon-lee (approved):
lgtm

cmd/doracled/cmd/gen_oracle_key.go
```diff
@@ -0,0 +1,112 @@
...
+	return nil
+}
+
+func getKeyPaths(enclaveEnabled bool) (string, string) {
+	if enclaveEnabled {
+		return types.DefaultOraclePrivKeyFilePath, types.DefaultOraclePubKeyFilePath
+	} else {
+		return filepath.Join(homeDir, types.DefaultOraclePrivKeyName), filepath.Join(homeDir, types.DefaultOraclePubKeyName)
+	}
+}
```
gyuguen: Is there a reason you separated this path?
I think it can be used in the same way in the enclave environment as below.
```go
func getKeyPaths() (string, string) {
    return filepath.Join(homeDir, types.DefaultOraclePrivKeyName), filepath.Join(homeDir, types.DefaultOraclePubKeyName)
}
```

Because it sets the HOME environment variable of encalve.json.

0xHansLee: You're right. Regardless of the environments, the `homeDir` is set properly. So we don't need to separate two cases.
I fixed it [5341f29](https://github.com/medibloc/panacea-doracle/pull/4/commits/5341f2945bd0d1ca50e44c1871636c92764df5f6)

gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM 👍

PR #8, opened 2022-07-28
feat: implementation of report verification

### LINKED ISSUE #6: Implement a cmd for remote report verification
Can verify the remote report using Azure DCAP client.

0xHansLee:
Close #6 

Implementation of `verify-report` cmd for verification of a remote report.
For verification, the promised enclave info(`ProductID`, `SignerID`, and `UniqueID`) to be compared is required.
They are retrieved from generating a self remote report instead of hardcoding of the values.

### TCBStatus

In verification of remote report (`sgx.VerifyRemoteReport`), there is no specific check for `TCBStatus`.
As I searched for `TCBStatus` or `TCBLevel`, it is related to the environment or configuration(BIOS) of the system that generates the remote report. But the `TCBStatus` is determined by Azure DCAP confidentially, no criteria for the status is known to outside. They say that it depends on service policy.

EGo returns an err(`ErrTCBLevelInvalid`) if the `TCBStatus` is not considered up-to-date. So I'd like to suggest to follow them for now (There are several statuses of up-to-date, but required some changes in config). If we need other policy of `TCBStatus`, then we can change.

Other changes
- `Enclave.enable` in config.toml is deleted. Accordingly, user can execute `gen-oracle-key` cmd only in SGX-enabled env.
- add suffix `Base64` to the fields in `OraclePubKeyInfo` since it is base64 encoded value.
- When generating remote report in `gen-oracle-key`, the additional data is changed to the hash value of public key.


README.md
```diff
@@ -141,6 +141,24 @@ Run the binary using `ego` so that it can be run in the secure enclave.
 AZDCAP_DEBUG_LOG_LEVEL=INFO ego run doracled gen-oracle-key
 ```
 
+### verify remote report
+
+You can verify that that key is generated in SGX using the promised binary.
```
gyuguen:
```suggestion
You can verify that key is generated in SGX using the promised binary.
```

cmd/doracled/cmd/gen_oracle_key.go
```diff
@@ -61,14 +55,15 @@ So please be cautious in using this command.`,
...
-		if err := sgx.SealToFile(oraclePrivKey.Serialize(), oraclePrivKeyPath, conf.Enclave.Enable); err != nil {
+		if err := sgx.SealToFile(oraclePrivKey.Serialize(), oraclePrivKeyPath); err != nil {
 			log.Errorf("failed to write %s: %v", oraclePrivKeyPath, err)
 			return err
 		}
 
 		// generate oracle key remote report
 		oraclePubKey := oraclePrivKey.PubKey().SerializeCompressed()
-		oracleKeyRemoteReport, err := sgx.GenerateRemoteReport(oraclePubKey[:], conf.Enclave.Enable)
+		oraclePubKeyHash := sha256.Sum256(oraclePubKey)
```
gyuguen: Is there a reason you hashed this?
Is it because of the data length limit of the report?

gyuguen (approved):
Thank you.
I left one suggested change and question.

cmd/doracled/cmd/gen_oracle_key.go
0xHansLee: Yes. However, since the length of pubkey is also fixed, we can use the pubkey as is.

0xHansLee: To make the `sgx.GenerateRemoteReport` more general, it can be improved as below,

```go
func GenerateRemoteReport(data []byte) ([]byte, error) {
        hashedData := sha256.Sum256(data)
	return enclave.GetRemoteReport(hashedData)
}
```

No matter how long the length of data is, this function can generate a valid remote report.

README.md
```diff
@@ -141,6 +141,24 @@ Run the binary using `ego` so that it can be run in the secure enclave.
...
+{
+  "public_key_base_64" : "<base64-encoded-public-key>",
+  "remote_report_base_64": "<base64-encoded-remote-report>"
+}
+```
+
+Then, you can verify the remote report.
+
+```bash
+AZDCAP_DEBUG_LOG_LEVEL=INFO ego run doracled verify-report <remote-report-path>
```
youngjoon-lee, line 159: Is this CLI for someone who wants to verify SGX reports uploaded to the Panacea if they want (for example, by a genesis oracle)?

0xHansLee: Exactly. Using this CLI, anyone can verify the report from genesis oracle.

youngjoon-lee: If so, it would be more suitable to modify the JSON file as below:
```json
{
  "remote_report_base64": "...",
  "remote_report_data_base64": "..."
}
```
, because remote reports can contain any data such as a public key or something else in the future.

cmd/doracled/cmd/gen_oracle_key.go
youngjoon-lee: IIRC, the remote report can contain max 512 bits. So, you can use sha512.

I would prefer executing hash functions outside of the `GenerateRemoteReport` explicitly.

README.md
youngjoon-lee: Ah. I think it’s not a big deal. Let’s keep going as you wish.

cmd/doracled/cmd/verify_report.go
```diff
@@ -0,0 +1,76 @@
...
+		return nil, err
+	}
+
+	return &pubKeyInfo, nil
+}
+
+func verifyPubKeyRemoteReport(pubKeyInfo OraclePubKeyInfo) error {
+	pubKey, err := base64.StdEncoding.DecodeString(pubKeyInfo.PublicKeyBase64)
+	if err != nil {
+		return fmt.Errorf("failed to decode oracle public key: %v", err)
```
youngjoon-lee:
```suggestion
		return fmt.Errorf("failed to decode oracle public key: %w", err)
```
Please wrap the error instead of just stringify it

sgx/types.go
```diff
@@ -0,0 +1,38 @@
...
+	if err != nil {
+		log.Errorf("failed to generate self-report: %v", err)
+		return nil, err
+	}
+
+	report, err := enclave.VerifyRemoteReport(reportBz)
+	if err != nil {
+		log.Errorf("failed to retrieve self-report: %v", err)
+		return nil, err
+	}
```
youngjoon-lee, line 35: Cuz you're verifying a report that was generated right before, this verification will be always successful. Then, what do you think about skipping this verification?

youngjoon-lee (approved):

sgx/types.go
0xHansLee: This is just to retrieve `attestation.Report` struct, not for verification. The only way to get structured data is from syscall, so I could get the structured data via `enclave.VerifyRemoteReport` which returns structured `attestation.Report` data.

Do you have any idea of getting `attestation.Report` struct from byte array?

The unique ID could be given as an argument of this CLI, but I and @gyuguen thought that it's the better way to generate self-remote report and compare with it.

0xHansLee: The reason why I'm asking is that I'm thinking about the better way to get unique ID as well as Product ID and signer ID. It works well, but little bit weird to me as well in that generating remote report and verifying it right after, as you said.

youngjoon-lee: Ah. Gotcha. Then, I think it's good to use `VerifyRemoteReport`.

cmd/doracled/cmd/gen_oracle_key.go
0xHansLee: Agreed. And I'm thinking not using hash of public key. There is no need to hash the public key because its length is shorter than _maxReportData_(64).

PR #10, opened 2022-08-08
feat: implementation of light client verification

audtlr24:
add `GetStoreData` for verification using light client and `ABIQueryWithOptions`.

`GetStoreData` get data from panacea and verify results using light client and merkle proof.
All query function that have to verify the results, can use `GetStoreData`.
`GetAccount`, `GetBalance`, `GetTopic` is an example of query that use `GetStoreData`.

* Note : `GetBalance` does not work because of current panacea cosmos-sdk version issue.

Other Change
* add `Seal`, `Unseal` to develop sgx level DB.
* add golevelDB codes at `/tm-db` to devleop sgx level DB.

Next Work
* change light client DB to sgx level DB


gyuguen:
Please chage the title of your PR
feat: xxx

panacea/grpc_client.go
```diff
@@ -32,7 +31,7 @@ func makeInterfaceRegistry() sdk.InterfaceRegistry {
 
 func NewGrpcClient(conf *config.Config) (GrpcClientI, error) {
 	log.Infof("dialing to Panacea gRPC endpoint: %s", conf.Panacea.GRPCAddr)
-	conn, err := grpc.Dial(conf.Panacea.GRPCAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
gyuguen, line 35: If this setting is removed, an error is likely to occur during SSL communication.
Would you like to test with https://testnet-grpc.gopanacea.org?

gyuguen: If there is no error, I think you can delete it.

panacea/query_client.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// GetBalance returns balance from address.
+func (c QueryClient) GetBalance(address string) (sdk.Coin, error) {
+	acc, err := sdk.GetFromBech32(address, "panacea")
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	var denom = "umed"
```
gyuguen: It would be better to subtract this value as const or receive it as a parameter.

panacea/query_client.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return sdk.Coin{}, err
+	}
+
+	return balance, nil
+}
+
+// GetTopic returns topic from address and topicName.
+func (c QueryClient) GetTopic(address string, topicName string) (aoltypes.Topic, error) {
+	acc, err := sdk.GetFromBech32(address, "panacea")
```
gyuguen: Let's merge these panacea-dependent things into panacea package later.

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+// a RpcClient that can be used to query data with light client verification & merkle proof.
+// Period of trustOptions can be changed according to slashing period.
+func NewRpcClient(ctx context.Context, chainID, rpcAddr string, trustedHeight int, trustedBlockHash []byte) (*RpcClient, error) {
+	rpcClient, err := rpchttp.New(rpcAddr, "/websocket")
+	if err != nil {
+		return nil, err
+	}
+
+	trustOptions := light.TrustOptions{
+		Period: 365 * 24 * time.Hour,
```
gyuguen: I think we should change this value to 2 weeks.
But first let's leave it like this

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+		Height: int64(trustedHeight),
+		Hash:   trustedBlockHash,
+	}
+
+	pv, err := http.New(chainID, rpcAddr)
+	if err != nil {
+		return nil, err
+	}
+	pvs := []provider.Provider{pv}
```
gyuguen: It is good to put information of other nodes in this value.
Putting our duplex in is one way, let's think about it more

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
+package panacea
+
+import (
+	"context"
+	"fmt"
+	ics23 "github.com/confio/ics23/go"
+	"github.com/cosmos/ibc-go/v2/modules/core/23-commitment/types"
```
gyuguen: I brought the logic to verify the merkle proof in the ibc module. This works fine, but I'm not sure if I should continue to use it from this module.
First, let's use it as is.

sgx/seal.go
```diff
@@ -22,3 +22,23 @@ func SealToFile(data []byte, filePath string) error {
 
 	return nil
 }
+
+// Seal returns data sealed with unique ID in SGX-enabled environments
+// If SGX is disabled, it returns the data as is.
+func Seal(data []byte, enclaveEnabled bool) ([]byte, error) {
+	if enclaveEnabled {
```
gyuguen: Perhaps the `enclaveEnabled` value will be deleted.

gyuguen:
There seems to be a lot of content in the current PR.
It would have been nice to have divided the LightClient part and the LevelDB part.

It's fine as it is for now.
I think it would be better to request a smaller PR in the future. 😄

panacea/grpc_client.go
0xHansLee: When you see an example given by cosmos-sdk ([link](https://github.com/cosmos/cosmos-sdk/blob/main/docs/run-node/interact-node.md#programmatically-via-go)), they said `The Cosmos SDK doesn't support any transport security mechanism.`. Please refer this example and check if there is no error with our testnet as @gyuguen said :)

youngjoon-lee:
> There seems to be a lot of content in the current PR. It would have been nice to have divided the LightClient part and the LevelDB part.
> 
> It's fine as it is for now. I think it would be better to request a smaller PR in the future. 😄

Yeah. We recommend to make PR as smaller as we can, so that reviewers can review a PR within 30m. As gyuguen said, it's good to have only one topic in a PR if possible. It's also good for fast feedback cycles.
In some cases, you need to implement bunch of codes at once in order to test complicated scenarios. Then, you can just open multiple PRs so that reviewers can understand what you've done step by step.

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/light"
+	"github.com/tendermint/tendermint/light/provider"
+	"github.com/tendermint/tendermint/light/provider/http"
```
0xHansLee: I think it'd be better to alias this `http` package to `tmhttp` or something like this for code readability.

panacea/rpc_client.go
0xHansLee: Agree

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+	result, err := q.RpcClient.ABCIQueryWithOptions(ctx, fmt.Sprintf("/store/%s/key", storeKey), key, option)
+	if err != nil {
+		return nil, err
+	}
+
+	// get trustedBlock at blockHeight+1
+	time.Sleep(blockPeriod) // AppHash for query is in the next block, so have to wait until the next block is confirmed
+
+	// wait a creation of the next block
+	textTrustedBlock, err := q.LightClient.VerifyLightBlockAtHeight(ctx, trustedBlock.Height+1, time.Now())
```
0xHansLee:
```suggestion
	nextTrustedBlock, err := q.LightClient.VerifyLightBlockAtHeight(ctx, trustedBlock.Height+1, time.Now())
```

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+
+type RpcClient struct {
+	RpcClient   *rpchttp.HTTP
+	LightClient *light.Client
+}
+
+// NewRpcClient set RpcClient with trustedBlock Info and returns, if successful,
+// a RpcClient that can be used to query data with light client verification & merkle proof.
+// Period of trustOptions can be changed according to slashing period.
+func NewRpcClient(ctx context.Context, chainID, rpcAddr string, trustedHeight int, trustedBlockHash []byte) (*RpcClient, error) {
```
0xHansLee: IMO, it would be good to merge the `QueryClient` with the `RpcClient` and name it `QueryClient`. Because,
- it is little bit confusing that `RpcClient` has a `RpcClient` which is a type of `*rpchttp.HTTP`.
- there are so many clients in this project, so let's remove unclear struct type.

In other words, the role of `QueryClient` is to get data from Panacea and validate it.

panacea/rpc_client.go
```diff
@@ -0,0 +1,113 @@
...
+		Height: int64(trustedHeight),
+		Hash:   trustedBlockHash,
+	}
+
+	pv, err := http.New(chainID, rpcAddr)
+	if err != nil {
+		return nil, err
+	}
+	pvs := []provider.Provider{pv}
+	store := dbs.New(dbm.NewMemDB(), chainID)
```
0xHansLee: To enable to seal and unseal the data, this store will be replaced to the `sgxLevelDB` you modified?

0xHansLee (commented):
Thank you for your efforts and I left some comments.

Btw, what cosmos-sdk version related issue do you have? The reason why test fail is due to the issue you mentioned? If then, you can comment the test for now. Or since the test is given as an example, so you can delete or replace it with other query.

And could you give a brief background or context about the sgx level db you developed? It's just addition of seal/unseal to Get/Set of db?

gyuguen:
> Thank you for your efforts and I left some comments.
> 
> Btw, what cosmos-sdk version related issue do you have? The reason why test fail is due to the issue you mentioned? If then, you can comment the test for now. Or since the test is given as an example, so you can delete or replace it with other query.
> 
> And could you give a brief background or context about the sgx level db you developed? It's just addition of seal/unseal to Get/Set of db?

The version issue of Cosmos-sdk is because the environment queried by the test code is mainnet, and the version of mainnet is cosmos-sdk 0.42.x.
The master branch of panacea-core is cosmos-sdk 0.45.x, so the storage method of the merkle tree is slightly different.
Perhaps he left that part as a comment.

audtlr24:
> Please chage the title of your PR feat: xxx

Thanks, I changed the title.

panacea/grpc_client.go
audtlr24: Thanks, I add `grpc.WithUnsecure()` to `Dial` Option. It is now work without error when I test with https://testnet-grpc.gopanacea.org . Please check the new commit.

panacea/query_client.go
audtlr24: Thanks, I reflect in a new commit.

panacea/query_client.go
audtlr24: I don't understand exactly, so can you explain in detail?

panacea/rpc_client.go
audtlr24: I agree. After add some logics (get initial trusted block info from user input, automatic update trusted block info from panacea), we can change this value to 2 weeks. 
For now, I set it to 2 years for testing.

panacea/rpc_client.go
audtlr24: I Agree. I will add that feature next time.

panacea/rpc_client.go
audtlr24: Thanks, I reflect in a new commit.

panacea/rpc_client.go
audtlr24: It's typo. Thanks for finding it.

panacea/rpc_client.go
audtlr24: Thanks, I reflect in a new commit. Please check the commit.

panacea/rpc_client.go
audtlr24: Yes, I will change this `dbm.NewMemDB()` to `sgxLevelDB`. Now, use tmp DB for test.

panacea/query_client.go
gyuguen: What I meant is that the logic related to panacea should be put in the panacea package.
Like this. (https://github.com/medibloc/panacea-coracle/tree/main/panacea)

audtlr24:
> Thank you for your efforts and I left some comments.
> 
> Btw, what cosmos-sdk version related issue do you have? The reason why test fail is due to the issue you mentioned? If then, you can comment the test for now. Or since the test is given as an example, so you can delete or replace it with other query.

@gyuguen explain the cosmos-sdk version issue well, thanks. 
I deleted the test `GetBalance` that caused the test error.

> 
> And could you give a brief background or context about the sgx level db you developed? It's just addition of seal/unseal to Get/Set of db?

Yes. the `sgxlevelDB` seal the value in key-value store. So I fix `put`, `get` and connected logic. 
However, I delete `sgxLevelDB` part in this PR. 
In the next PR, I will be able to put more explanations.

panacea/query_client.go
gyuguen: Oh by the way, it's already in the panacea package. :)

gyuguen: So, what if you create an `account.go` file and make it implements like this:
Because I think It will be able to use these address converters often.

account.go
```go
const prefix = "panacea"

func GetAccAddressFromBech32(address string) (addr sdk.AccAddress, err error) {
  return sdk.GetFromBech32(address, prefix)
}
```

panacea/query_client.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return aoltypes.Topic{}, err
+	}
+
+	key := aoltypes.TopicCompositeKey{OwnerAddress: acc, TopicName: topicName}
+	topicKey := append(aoltypes.TopicKeyPrefix, compkey.MustEncode(&key)...)
+	bz, err := c.GetStoreData(context.Background(), aoltypes.StoreKey, topicKey)
```
gyuguen, line 188: I think the error handling is missing.

go.mod
```diff
@@ -4,13 +4,18 @@ go 1.17
 
 require (
 	github.com/btcsuite/btcd v0.22.1
-	github.com/cosmos/cosmos-sdk v0.45.1
+	github.com/confio/ics23/go v0.7.0
+	github.com/cosmos/cosmos-sdk v0.45.4
+	github.com/cosmos/ibc-go/v2 v2.0.3
 	github.com/edgelesssys/ego v1.0.0
+	github.com/medibloc/panacea-core/v2 v2.0.4
```
gyuguen: What about using the current master version of panacea-core?
```
go get github.com/medibloc/panacea-core/v2@master
```

panacea/query_client.go
audtlr24: Thanks, I agree. I reflect in the commit [60b9447](https://github.com/medibloc/panacea-doracle/pull/10/commits/60b94479345c6386c8704ce1d15fe3612605900e).

panacea/query_client.go
audtlr24: Oh thank you, now it is fixed.

go.mod
audtlr24: Thanks, I changed it at [60b9447](https://github.com/medibloc/panacea-doracle/pull/10/commits/60b94479345c6386c8704ce1d15fe3612605900e).

audtlr24:
> There seems to be a lot of content in the current PR. It would have been nice to have divided the LightClient part and the LevelDB part.
> 
> It's fine as it is for now. I think it would be better to request a smaller PR in the future. 😄



> > There seems to be a lot of content in the current PR. It would have been nice to have divided the LightClient part and the LevelDB part.
> > It's fine as it is for now. I think it would be better to request a smaller PR in the future. 😄
> 
> Yeah. We recommend to make PR as smaller as we can, so that reviewers can review a PR within 30m. As gyuguen said, it's good to have only one topic in a PR if possible. It's also good for fast feedback cycles. In some cases, you need to implement bunch of codes at once in order to test complicated scenarios. Then, you can just open multiple PRs so that reviewers can understand what you've done step by step.

@gyuguen @youngjoon-lee Thanks for your comments. So, I delete the `sgxLevelDB` part in this PR. I'll post it in the next PR after making some corrections.

panacea/query_client.go
gyuguen: The commit you shared does not reflect any modifications to the content.

gyuguen: I found this commit. :)
https://github.com/medibloc/panacea-doracle/pull/10/commits/75b1490b373ee04860d6a91cb5288f47b51a6400

gyuguen (approved):
LGTM 👍

panacea/rpc_client.go
0xHansLee: I got it. Thanks for comment!

0xHansLee (approved):
LGTM

youngjoon-lee (commented):
lgtm

PR #11, opened 2022-08-11
feat: implementation of register-oracle cmd

0xHansLee:
Implementation of `register-oracle` command.
The information of trusted block is required flag of this command, which is used for light client.
I successfully tested the functionality with interaction of Panacea under SGX-enabled env locally.
(I also confirmed that if oracle request the registration with different mnemonic which is not registered as validator, the tx fails)

This PR includes simple functionality, not error-handling, graceful shutdown, and so on.
Some considerations of several cases (e.g. success generating node key, but stop process, and then restart) should be handled as well, including event subscriber. 
I'll make those in the next PR. Some refactoring of structure might be included in the next PR, but if you have any comments, feel free to leave your thoughts.


gyuguen (approved):
LGTM 👍

audtlr24 (approved):
LGTM!

config/config.go
```diff
@@ -24,13 +29,19 @@ func DefaultConfig() *Config {
 			ListenAddr:     "127.0.0.1:8080",
 		},
 		Panacea: PanaceaConfig{
-			GRPCAddr: "127.0.0.1:9090",
+			ChainID:          "",
+			GRPCAddr:         "127.0.0.1:9090",
+			DefaultGasLimit:  300000,
+			DefaultFeeAmount: "1000000umed",
```
youngjoon-lee: Would be set `DefaultGasLimit` to `200000` if you're gonna set `DefaultFeeAmount` to `1000000umed`, because our recommended gas price is `5umed`.

panacea/consts.go
```diff
@@ -0,0 +1,6 @@
+package panacea
+
+const (
+	CoinType = 371
+	Hrp      = "panacea"
```
youngjoon-lee: it is recommeneded to use uppercase `HRP` for abbreviations.

config/config.go
0xHansLee: Right. I changed `DefaultGasLimit` to 200000 in [5a20456](https://github.com/medibloc/panacea-doracle/pull/11/commits/5a2045637c4e4834e5cf21efb76488f73152fd0c)

panacea/consts.go
0xHansLee: I got it! Thanks. fixed in [5a20456](https://github.com/medibloc/panacea-doracle/pull/11/commits/5a2045637c4e4834e5cf21efb76488f73152fd0c)

inchori (approved):
LGTM 👍

PR #12, opened 2022-08-22
feat: Implementation of conversion to sgx encrypted levelDB for light client

audtlr24:
add sgxLevelDB for storing encrypted light block to level DB (using seal, unseal function).

`/store/sgxLevelDB` is the levelDB using `Seal` and `Unseal` functions when `Set` and `Get` the value from levelDB. 
I replaced`query_client` DB from `NewMemDB()` to `NewGolevelDB()` that  uses sgx `UniqueKey`.
The levelDB path can be set in the `query_client`.

Note : All the tests using light client, including `TestGetAccount` and `TestSgxLevelDB`, can only work in sgx environment, so the tests are commented out. If you want test, uncomment it and test it.


gyuguen:
https://go.dev/blog/package-names

In go, package names should be short and lowercase.

How about changing the pacakge name below?
`store/sgxLevelDB` -> `store/leveldb`

And I would like to change the file name as well.
```
sgxLeveldb.go -> sgx.go
sgxleveldb_batch.go -> sgx_batch.go
sgxleveldb_iterator.go -> sgx_iterator.go
```

panacea/query_client.go
```diff
@@ -52,7 +54,20 @@ func NewQueryClient(ctx context.Context, chainID, rpcAddr string, trustedBlockHe
...
 	}
 	pvs := []provider.Provider{pv}
-	store := dbs.New(dbm.NewMemDB(), chainID)
+
+	if _, err := os.Stat(leveldbPath); os.IsNotExist(err) {
+		err = os.MkdirAll(leveldbPath, 0755)
+		if err != nil {
+			return nil, err
+		}
+	}
```
gyuguen: Does this work well?
My guess is that the sgx execution environment won't find this file path.

The executable path in our sgx environment is:
`/data`
However, the path referenced here now seems to go down one level than `/data`.

You can refer to [this code](https://github.com/medibloc/panacea-doracle/blob/main/cmd/doracled/cmd/start.go#L17) to see how to read `config.toml`

Finally, it would be good if this data is saved in the following path in sgx environment.
```
$HOME/.doracle/data
```

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+
+var backends = map[tmdb.BackendType]dbCreator{}
+
+func registerDBCreator(backend tmdb.BackendType, creator dbCreator, force bool) {
+	_, ok := backends[backend]
+	if !force && ok {
+		return
+	}
+	backends[backend] = creator
+}
```
gyuguen: Are these operations necessary to initialize leveldb?

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Get implements DB.
+func (db *GoLevelDB) Get(key []byte) ([]byte, error) {
+	if len(key) == 0 {
+		return nil, errKeyEmpty
+	}
+	res, err := db.Db.Get(key, nil)
+	if err != nil {
+		if err == errors.ErrNotFound {
+			return nil, nil
```
gyuguen: Is it nil, not an error because there is no data?

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+	return unsealedRes, nil
+}
+
+// Has implements DB.
+func (db *GoLevelDB) Has(key []byte) (bool, error) {
+	bytes, err := db.Get(key)
+	if err != nil {
+		return false, err
+	}
+	return bytes != nil, nil
```
gyuguen: think it can be easily modified like this.
```suggestion
	return db.DB.Has(key)
```

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+
+// Delete implements DB.
+func (db *GoLevelDB) Delete(key []byte) error {
+	if len(key) == 0 {
+		return errKeyEmpty
+	}
+	if err := db.Db.Delete(key, nil); err != nil {
+		return err
+	}
+	return nil
```
gyuguen:
```suggestion
	return db.Db.Delete(key, nil)
```

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+// DeleteSync implements DB.
+func (db *GoLevelDB) DeleteSync(key []byte) error {
+	if len(key) == 0 {
+		return errKeyEmpty
+	}
+	err := db.Db.Delete(key, &opt.WriteOptions{Sync: true})
+	if err != nil {
+		return err
+	}
+	return nil
```
gyuguen:
```suggestion
	return db.Db.Delete(key, &opt.WriteOptions{Sync: true})

```

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+func (db *GoLevelDB) DB() *leveldb.DB {
+	return db.Db
+}
+
+// Close implements DB.
+func (db *GoLevelDB) Close() error {
+	if err := db.Db.Close(); err != nil {
+		return err
+	}
+	return nil
```
gyuguen:
```suggestion
	return db.Db.Close()
```

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
...
+		if err != nil {
+			return err
+		}
+		fmt.Printf("[%X]:\t[%X]\n", key, unsealValue)
+	}
+	return nil
+}
+
+// Stats implements DB.
+func (db *GoLevelDB) Stats() map[string]string {
```
gyuguen: What did you refer to this?

gyuguen (commented):
Thank you.
I left a few comment.

gyuguen:
What did `sgxleveldb_batch.go` and `sgxleveldb_iterator.go` refer to?

audtlr24:
> What did `sgxleveldb_batch.go` and `sgxleveldb_iterator.go` refer to?

It required for light client internal logics. For example, `SaveLightBlock` uses database batch to set the data. And `LastLightBlockHeight` uses database iterator. You can refer this at [here](https://github.com/tendermint/tendermint/blob/v0.34.19/light/store/db/db.go).

store/sgxLevelDB/sgxLeveldb.go
audtlr24: Thanks, I think your suggestions are reasonable. But, I copy the code from [tendermint/tm-db](https://github.com/tendermint/tm-db/blob/v0.6.7/goleveldb.go), so It is difficult to decide whether it is better to modify it or follow it as it is.

What about your opinion? @gyuguen

gyuguen: Aha. Are all of them from tm-db except for the sgx use part?

store/sgxLevelDB/sgxLeveldb.go
```diff
@@ -0,0 +1,251 @@
+// Package sgxLevelDB implements a light client level db for panacea-doracle.
+// It does include Set & Get functions that are sealed & unsealed in the sgx environment.
+
+package sgxLevelDB
```
youngjoon-lee: pls rename this package to `sgxleveldb`.
https://go.dev/doc/effective_go#package-names

store/sgxLevelDB/sgxLeveldb.go
audtlr24: Yes, I just fix some sgx use part.

store/sgxLevelDB/sgxLeveldb.go
gyuguen: Let's change the package name as @youngjoon-lee said. :)

gyuguen:
Well, I'm still thinking about the package name and the file name.
I think it would be good to go like below...🥲
```
store
  ㄴsgxleveldb
    ㄴdb.go
    ㄴdb-batch.go
    ㄴdb-iterator.go
```

store/sgxLevelDB/sgxLeveldb.go
gyuguen: Does the read/write function with LevelDB work well in the SGX environment?
Then I see no problem.

audtlr24: Yes, it works well. I perform test that reads the value stored in sgxlevelDB directly, unseals it, and compares it with the original value.

panacea/query_client_test.go
```diff
@@ -1,34 +1,28 @@
...
 
-import (
-	"context"
-	"encoding/hex"
-	"github.com/cosmos/cosmos-sdk/types/bech32"
-	"github.com/medibloc/panacea-doracle/panacea"
-	"github.com/stretchr/testify/require"
-	"testing"
-)
+// All the tests can only work in sgx environment, so the tests are commented out.
```
youngjoon-lee, line 3: Okay. I think it's okay to assume that this project works only in SGX (It means that you don't need to care about the local dev on MacOS). But, let's go with commenting this out for now. I'll reoragnize the whole testing strategy soon.

sgx/seal.go
```diff
@@ -22,3 +22,23 @@ func SealToFile(data []byte, filePath string) error {
 
 	return nil
 }
+
+// Seal returns data sealed with unique ID in SGX-enabled environments
+// If SGX is disabled, it returns the data as is.
+func Seal(data []byte, enclaveEnabled bool) ([]byte, error) {
```
youngjoon-lee, line 42: I don't think it's good to have the `enclaveEnabled` param here. As I mentioned [above](https://github.com/medibloc/panacea-doracle/pull/12#discussion_r951189311), I think we don't need to make this project work in non-SGX environments.

store/sgxleveldb/sgx.go
```diff
@@ -0,0 +1,251 @@
...
+		return nil, err
+	}
+	// unseal data
+	unsealedRes, err := sgx.Unseal(res, true)
+	log.Debugf("Unseal value from key : %X\n", key)
+	if err != nil {
+		return nil, err
+	}
+
+	return unsealedRes, nil
```
0xHansLee: You can just `return sgx.Unseal(res, true)`
```suggestion
	return sgx.Unseal(res, true)
```

panacea/query_client.go
audtlr24: Thanks, I fixed it at https://github.com/medibloc/panacea-doracle/pull/12/commits/6fcd8d0559e59a7a72376281d104d78957840d43.
It would be better to change to put that `dbDir` in the config.
I will reflect it in the next PR.

sgx/seal.go
youngjoon-lee: Let's have this as you did for now. I will check how we can improve our CI environment.

store/sgxleveldb/sgx.go
```diff
@@ -0,0 +1,251 @@
...
+
+var backends = map[tmdb.BackendType]dbCreator{}
+
+func registerDBCreator(backend tmdb.BackendType, creator dbCreator, force bool) {
+	_, ok := backends[backend]
+	if !force && ok {
+		return
+	}
+	backends[backend] = creator
+}
```
youngjoon-lee: It seems that the `backends` map isn't used anywhere after being initialized.

store/sgxleveldb/sgx.go
```diff
@@ -0,0 +1,251 @@
...
+
+type GoLevelDB struct {
+	Db *leveldb.DB
+}
+
+var _ tmdb.DB = (*GoLevelDB)(nil)
+
+func NewGoLevelDB(name string, dir string) (*GoLevelDB, error) {
+	return NewGoLevelDBWithOpts(name, dir, nil)
+}
```
youngjoon-lee: In Go, it's more recommended to use the interface `tmdb.DB` as a return type, if you don't have any specific reason to return the `*GoLevelDB` explicitly. It's for maximizing the power of interfaces.

Then, you can probably simplify your `init()` above.
```go
func init() {
    registerDBCreator(tmdb.GoLevelDBBackend, NewGoLevelDB, false)
}
```

youngjoon-lee: But, this isn't a big deal now. Please resolve the https://github.com/medibloc/panacea-doracle/pull/12#discussion_r951207998 first.

store/sgxleveldb/sgx.go
youngjoon-lee: Yes, if you aren't gonna wrap the error: https://trstringer.com/errors-and-error-wrapping-go/.

store/sgxleveldb/sgx.go
```diff
@@ -0,0 +1,251 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return unsealedRes, nil
+}
+
+// Has implements DB.
+func (db *GoLevelDB) Has(key []byte) (bool, error) {
+	bytes, err := db.Get(key)
```
youngjoon-lee: I guess you can just `db.Db.Has(key)`.

store/sgxleveldb/sgx.go
youngjoon-lee: Also, you don't needa put `\n` in your log message. The log package (or logrus package) does it for you.

0xHansLee: Ah, right. I think it would be good to wrap the error from unseal. Thank you!

audtlr24: Thanks, I delete `\n` in log message.

youngjoon-lee:
@audtlr24 @gyuguen I think we can inherit the `tmdb.GoLevelDB` using Go's [struct embedding](https://gobyexample.com/struct-embedding). Then, we can override only functions that we want to override.
https://gist.github.com/youngjoon-lee/8bb61174ec07e57e6baf0f1396d04033

I haven't tested this code with your codebase, but I guess it will work.

store/sgxleveldb/sgx.go
youngjoon-lee: This topic has been moved to https://github.com/medibloc/panacea-doracle/pull/12#issuecomment-1223638247

store/sgxleveldb/sgx.go
youngjoon-lee: This topic has been moved to https://github.com/medibloc/panacea-doracle/pull/12#issuecomment-1223638247

store/sgxleveldb/sgx.go
youngjoon-lee: This topic has been moved to https://github.com/medibloc/panacea-doracle/pull/12#issuecomment-1223638247

gyuguen:
> @audtlr24 @gyuguen I think we can inherit the `tmdb.GoLevelDB` using Go's [struct embedding](https://gobyexample.com/struct-embedding). Then, we can override only functions that we want to override. https://gist.github.com/youngjoon-lee/8bb61174ec07e57e6baf0f1396d04033
> 
> I haven't tested this code with your codebase, but I guess it will work.

Good!

audtlr24:
> 

Thanks, I will try it to the code, and test it.

audtlr24:
@gyuguen @youngjoon-lee @H4NLee Some of the code has been fixed in this PR. sgxLevelDB is inherited using Go's struct embedding and I set some parameters of `NewQueryClient` to `config.toml`. Please check again and comment.

config/config.go
```diff
@@ -19,6 +19,9 @@ type PanaceaConfig struct {
 	GRPCAddr         string `mapstructure:"grpc-addr"`
 	DefaultGasLimit  uint64 `mapstructure:"default-gas-limit"`
 	DefaultFeeAmount string `mapstructure:"default-fee-amount"`
+	PrimaryAddr   string `mapstructure:"primary-addr"`
+	WitnessesAddr string `mapstructure:"witnesses-addr"`
```
gyuguen, line 23: This should probably be able to accept multiple values.(separated by `,`)
Because `Witnessess` can receive multiple addresses

config/config.go
```diff
@@ -29,10 +32,13 @@ func DefaultConfig() *Config {
 			ListenAddr:     "127.0.0.1:8080",
 		},
 		Panacea: PanaceaConfig{
-			ChainID:          "",
+			ChainID:          "panacea-3",
 			GRPCAddr:         "127.0.0.1:9090",
```
gyuguen:
```suggestion
			GRPCAddr:         "https://grpc.gopanacea.org:443",
```

gyuguen: It would be good to unify the address on the mainnet or local

config/config.go
gyuguen: Ah, you have already set it to separate with whitespace. I think it would be better to separate them with a comma. :)

store/sgxleveldb/db.go
```diff
@@ -0,0 +1,57 @@
+// Package leveldb implements a light client level db for panacea-doracle.
```
gyuguen, line 1: This part is very simple :)

config/config.go
audtlr24: Thanks, I fix it in [da46b93](https://github.com/medibloc/panacea-doracle/pull/12/commits/da46b93be929962bac9329a82cf37c1e408b84d4).

gyuguen (approved):
LGTM 👍

panacea/query_client.go
```diff
@@ -40,8 +44,9 @@ type QueryClient struct {
 
 // NewQueryClient set QueryClient with rpcClient & and returns, if successful,
 // a QueryClient that can be used to add query function.
-func NewQueryClient(ctx context.Context, chainID, rpcAddr string, trustedBlockHeight int, trustedBlockHash []byte) (*QueryClient, error) {
-	rpcClient, err := rpchttp.New(rpcAddr, "/websocket")
+func NewQueryClient(ctx context.Context, config *config.Config, trustedBlockHeight int, trustedBlockHash []byte) (*QueryClient, error) {
```
0xHansLee: You can use `TrustedBlockInfo` struct in line 34 of this file :)

BTW, when creating new query client, can the trusted block info be retrieved from sgx level DB that you developed? not as args.

audtlr24: Oh, thanks. I will change to use `TrustedBlockInfo`.

There is a part that checks whether the light block is stored before in the levelDB when `light.NewClient` is called. So, it can be retrieved. I test it in the sgx environment as well. Does this answer the question right?

audtlr24: I fixed it in [8d45752](https://github.com/medibloc/panacea-doracle/pull/12/commits/8d4575274a8eaf1656b1f5e99f7364dcaca8d27f).

0xHansLee: As I understand, the trusted block info is given by user only when registering oracle. 
At this time, the trusted block info should be sealed and stored to be utilized for light client. (I also missed to implement this part)

That is, the trusted block info should only be managed by the light client with sgx level DB after registration, not user.
If the trusted block info is required when process starts, the oracle should get the trusted block info from the sgx level DB. Because the trusted block info must not be changed arbitrarily.

audtlr24: @H4NLee As we said offline, I implement functions to store and retrieve trusted block info in levelDB. It is implemented in `panacea/sgxdb_util.go`. Thank you for checking again.

0xHansLee:
I think we need a `Close` function of the query client to close the connection.

audtlr24:
> I think we need a `Close` function of the query client to close the connection.

Okay, I see. Is it necessary to `Close` and `reconnect` the light client because cmd is split?
I add `Close()` in [5d9c394](https://github.com/medibloc/panacea-doracle/pull/12/commits/5d9c39408828f0f96e936a23b88901724631c3d7). If you call `NewQueryClient` again without `Close()` , an error occurs in the levelDB connection, which can be resolved by `Close()`.

panacea/sgxdb_util.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+	dbDir := filepath.Join(userHomeDir, ".doracle", "data")
+
+	db, err := sgxleveldb.NewSgxLevelDB("light-client-db", dbDir)
+	if err != nil {
+		return err
+	}
+	err = db.Set([]byte("trustedBlockInfo"), buffer.Bytes())
```
0xHansLee: Is the key `[]byte("trustedBlockInfo")` the same key used to store and retrieve the trusted block info by light client?

panacea/sgxdb_util.go
```diff
@@ -0,0 +1,83 @@
...
+	err := enc.Encode(info)
+	if err != nil {
+		return err
+	}
+
+	userHomeDir, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	dbDir := filepath.Join(userHomeDir, ".doracle", "data")
```
0xHansLee: This `dbDir` seems to be defined and used in several places. Why don't you define the `dbDir` as a global var and reuse it?

panacea/sgxdb_util.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	err = db.Set([]byte("trustedBlockInfo"), buffer.Bytes())
+	if err != nil {
+		return err
+	}
+
+	err = db.Close()
+	if err != nil {
+		return err
+	}
```
0xHansLee: Why don't you call the `db.Close()` in defer function?

panacea/sgxdb_util.go
audtlr24: Light client does not store `trustedBlockInfo` directly. It just stores `LightBlock` that has signed header and validator set. So, I saved the data with a new key `trustedBlockInfo`.

0xHansLee: Ah, I got it. Then, we need to update this `trustedBlockInfo` when trusted `LightBlock` is updated as well?

audtlr24: I think it's not necessary. The `trustedBlockInfo` is used to initialize and get `LightBlock` from primary. Then, for validation or update, the `Lightblock` is used and `trustedBlockInfo` is not.

0xHansLee: Understood.

I think that when `register-oracle` cmd is executed, the trusted block info should be set using the `SaveTrustedBlockInfo` function. 
Then, when `get-oracle-key` to be shared the oracle key, a query client is created with the trusted block info using `GetTrustedBlockInfo` function. 
After this initialization of query client (which includes light client), the light client will update and use the `LightBlock` as you said.

audtlr24: You're right. Good!

panacea/sgxdb_util.go
audtlr24: Thanks, I fixed it in [c003735](https://github.com/medibloc/panacea-doracle/pull/12/commits/c00373570e183550673268dde1e2e71c7958e956).

panacea/sgxdb_util.go
audtlr24: Thanks, I fixed it in [c003735](https://github.com/medibloc/panacea-doracle/pull/12/commits/c00373570e183550673268dde1e2e71c7958e956).

panacea/sgxdb_util.go
```diff
@@ -0,0 +1,76 @@
+package panacea
```
0xHansLee: Why don't we move this util functions to `sgxleveldb` package? Because it is more related to the db stuffs.

panacea/sgxdb_util.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		panic(err)
+	}
+	dbDir = filepath.Join(userHomeDir, ".doracle", "data")
+}
+
+func SaveTrustedBlockInfo(info TrustedBlockInfo) error {
+	var buffer bytes.Buffer
+
+	enc := gob.NewEncoder(&buffer)
```
0xHansLee: May I ask why do you use `encoding/gob` package to encode/decode the data?

cmd/doracled/cmd/init.go
```diff
@@ -26,6 +26,14 @@ var initCmd = &cobra.Command{
...
 		}
 
+		dbDir := filepath.Join(homeDir, "data")
+		if _, err := os.Stat(dbDir); os.IsNotExist(err) {
+			err = os.MkdirAll(dbDir, 0755)
+			if err != nil {
+				return fmt.Errorf("failed to create db dir: %w", err)
+			}
+		}
+
```
0xHansLee: `dbDir` is also used here :)

0xHansLee (approved):
utACK

left minor comments. Thank you for your works!

panacea/sgxdb_util.go
audtlr24: I'm not sure how to encode/decode struct to byte array, so I use `encoding/gob` package. Do you have any better way?

panacea/sgxdb_util.go
audtlr24: The `sgxdb_util.go` use `panacea` package for get `TrustedBlockInfo` structure. And panacea package use `sgxleveldb` package too. So when I move this functions to `sgxleveldb`, it throws `import cycle not allowed` error.

0xHansLee: I see. Then, let's keep this way

panacea/sgxdb_util.go
0xHansLee: I'm not sure too which one is the best way to us to encode/decode data, but there are several ways, for example `json.Marshal`/`json.Unmarshal`. So I just wonder the reason why you chose this package. I think there is no problem :)

audtlr24: Thanks, I will compare it 👍 .

cmd/doracled/cmd/init.go
audtlr24: Thanks, I fixed it in [30a6894](https://github.com/medibloc/panacea-doracle/pull/12/commits/30a6894d0df2bd7366209bbf994c493b6537ffdf).

panacea/sgxdb_util.go
0xHansLee: If the trusted block info given by operator(person) is only used to initialize the light client, actually we don't have to utilize sgx level DB. We can just seal and store the trusted block info as a file(e.g. .json). wdyt?

At first, I misunderstood that the light client uses and updates the trusted block info, so I said that this util functions are needed.

audtlr24: Yes.. I agree with your opinion. There seems to be no reason to use levelDB.

0xHansLee: OK. Then, let's store it as a file, not using level DB.
You can remove the sgxdb_util.go file.. sorry for the confusion

audtlr24: OK! 
No. I didn't even think deeply :).

PR #13, opened 2022-08-22
feat: re-structuring of register-oracle cmd

### LINKED ISSUE #7: Implement a cmd to request MsgRegisterOracle to Panacea
The followings need to be implemented:

- generating node key for sharing of oracle key
- sealing and storing the generated node private key
- generating remote report of the key
- broadcasting the MsgRegisterOracle Tx to Panacea

0xHansLee:
Close #7 

Some refactoring has been done including graceful shutdown and operation for each situation.

First, check if there is an existing node key.

If exists, continue the oracle registration process with the existing node key.
Or create a new one and request oracle registration to Panacea (done in #11).

When the node key exists, some checks would be done as below. 
- check if the existing node public key is equal to the one used in `OracleRegistration` in Panacea
- check the status of `OracleRegistration` (handling for each situation: `VotingPeriod`, `Passed`, `Rejected`)

### DIFF
--- a/cmd/doracled/cmd/register_oracle.go
+++ b/cmd/doracled/cmd/register_oracle.go
@@ -1,27 +1,16 @@
 package cmd
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
-	sdk "github.com/cosmos/cosmos-sdk/types"
-	"github.com/edgelesssys/ego/enclave"
-	oracletypes "github.com/medibloc/panacea-core/v2/x/oracle/types"
+	"github.com/medibloc/panacea-doracle/client/flags"
 	"github.com/medibloc/panacea-doracle/config"
-	"github.com/medibloc/panacea-doracle/crypto"
-	"github.com/medibloc/panacea-doracle/panacea"
-	"github.com/medibloc/panacea-doracle/sgx"
-	"github.com/medibloc/panacea-doracle/types"
-	log "github.com/sirupsen/logrus"
+	"github.com/medibloc/panacea-doracle/server/regoracle"
+	"github.com/medibloc/panacea-doracle/server/service"
 	"github.com/spf13/cobra"
-	"path/filepath"
-)
+	"os"
+	"os/signal"
 
-const (
-	FlagTrustedBlockHeight = "trusted-block-height"
-	FlagTrustedBlockHash   = "trusted-block-hash"
-	FlagAccNum             = "acc-num"
-	FlagIndex              = "index"
+	log "github.com/sirupsen/logrus"
 )
 
 func RegisterOracleCmd() *cobra.Command {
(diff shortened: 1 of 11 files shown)


server/service/service.go
```diff
@@ -1,30 +1,98 @@
...
 	Conf          *config.Config
-	Store         store.Storage
+	OracleAccount *panacea.OracleAccount
+	TrustedHeight int64
+	TrustedHash   []byte
 	PanaceaClient panacea.GrpcClientI
 }
 
-func New(conf *config.Config) (*Service, error) {
+func New(cmd *cobra.Command, conf *config.Config) (*Service, error) {
```
gyuguen: How about moving this cmd variable to cmd package as before?

It would be better to pass the values (trustedBlockHeight, trustedBlockHash) obtained here rather than the command object. Or make a struct and put it there and pass it on.

cmd/doracled/cmd/start.go
```diff
@@ -23,7 +22,7 @@ var startCmd = &cobra.Command{
 			return fmt.Errorf("failed to init logger: %w", err)
 		}
 
-		return server.Run(conf)
+		return nil
```
gyuguen, line 25: Is this a temporary declaration of nil?

panacea/grpc_client.go
```diff
@@ -81,6 +83,23 @@ func (c *GrpcClient) GetAccount(panaceaAddr string) (authtypes.AccountI, error)
...
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+	defer cancel()
+
+	reqMsg := &oracletypes.QueryOracleRegistrationRequest{
+		UniqueId: uniqueID,
+		Address:  oracleAddr,
+	}
+	response, err := client.OracleRegistration(ctx, reqMsg)
+	if err != nil {
+		return &oracletypes.OracleRegistration{}, err
```
gyuguen: I think it would be good for OracleRegistration to return nil when an error occurs.
In general, when an error occurs, it is thought that only the error is checked and not the object that responded. :)

```suggestion
		return nil, err
```

server/regoracle/server.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func (srv *Server) Close() error {
+	return nil
+}
+
+// continueRegistration continues the oracle registration process using the existing node key
+func (srv *Server) continueRegistration() error {
+	// get existing node key
+	nodePrivKeyBz, _ := sgx.UnsealFromFile(nodePrivKeyPath)
```
gyuguen: Wouldn't error checking be necessary here?

gyuguen (commented):
Thank you!
I left a few comment.

server/service/service.go
0xHansLee: I see. I agree with you. I'll make a new struct for the trusted block info.

cmd/doracled/cmd/start.go
0xHansLee: Yes. We have not implement the `start` cmd yet, so I just make it return nil.

server/regoracle/server.go
0xHansLee: Oh, I missed. I'll fix it

panacea/grpc_client.go
0xHansLee: ty. I changed it in [c9e91fc](https://github.com/medibloc/panacea-doracle/pull/13/commits/c9e91fc4af0152cff0421497127f61ff92c3ddf0)

server/regoracle/server.go
0xHansLee: fixed in [c9e91fc](https://github.com/medibloc/panacea-doracle/pull/13/commits/c9e91fc4af0152cff0421497127f61ff92c3ddf0)

server/service/service.go
0xHansLee: I changed in [adda53a](https://github.com/medibloc/panacea-doracle/pull/13/commits/adda53a5c1dbf4220d0dca7e6099a0f0bd1ba9b4) as you suggested. Thanks!

gyuguen (approved):
LGTM 👍

gyuguen:
Please add another reviewers :)

cmd/doracled/cmd/register_oracle.go
```diff
@@ -32,110 +23,120 @@ func RegisterOracleCmd() *cobra.Command {
...
-
-			cli, txBuilder, err := generateGrpcClientAndTxBuilder(conf)
-			if err != nil {
-				log.Errorf("failed to generate gRPC client and/or Tx builder: %v", err)
-				return err
+			defer svc.Close()
+
+			errChan := make(chan error, 1)
+
+			srv := regoracle.NewServer(svc)
```
youngjoon-lee, line 58: I don't get why this CLI needs a server and graceful shutdown. This makes things complicated.
Also the thing that you implemented is not a graceful shutdown.
Let's have a talk today.

youngjoon-lee (changes requested):

server/regoracle/server.go
```diff
@@ -0,0 +1,171 @@
...
+	// get unique ID
+	selfEnclaveInfo, err := sgx.GetSelfEnclaveInfo()
+	if err != nil {
+		return err
+	}
+
+	uniqueID := base64.StdEncoding.EncodeToString(selfEnclaveInfo.UniqueID)
+
+	// get oracle registration from Panacea.
+	// TODO: Should be replaced with light client query for data verification
```
audtlr24, line 67: Thanks for marking TODO. 
I think we initialize query client with `NewQueryClient` at `query_client.go` , before send query to panacea.
At this time, trusted block info, context, rpcAddr and chainId should be needed. So, I'm wondering where is the best place to call `NewQueryClient`.  Also, we should consider shut down of the query client  if oracle registration fails.

What about your opinion?

0xHansLee: The trusted block info would be given via flag of this command(`FlagTrustedBlockHeight`, `FlagTrustedBlockHash`).
And rpcAddr and chainID are added in `config.toml` as configuration params.

With our current structure, I would create a `NewQueryClient` inside `service.New` and the service would have a query client just like a gRPC client.

0xHansLee:
I would close this PR since we decided to make this `register-oracle` cmd simple.
This command will only execute generating node key and submitting transaction for registration of oracle to Panacea.

PR #15, opened 2022-08-23
feat: implement an event subscriber for RegisterOracle

### LINKED ISSUE #9: Implement an Event Interface and Subscriber
Implement an `Event` Interface. The `Event` Interface has the following method:
- `GetEventType()`
- `GetEventAttribute()`
- `GetEventHandler()`

Implement a `Subscriber`.
- When doracle starts, the doracle run subscriber with Tendermint RPC.
- The Tendermint websocket RPC address will be configured in `config.go`.

Implement a `RegisterOracle` event which implements an `Event` interface.
- Subscribe a `RegisterOracle` event.

inchori:
- Implement an event subscriber for RegisterOracle.
- The EventType, EventAttributeKey, and EventAttributeValue of `RegisterOracle` is `message.action='RegisterOracle'`.
- For now, subscriber running when `doracled start`.


config/config.go
```diff
@@ -22,9 +24,11 @@ func DefaultConfig() *Config {
 			LogLevel:       "info",
 			OracleMnemonic: "",
 			ListenAddr:     "127.0.0.1:8080",
+			Subscriber:     "",
```
gyuguen: What is a `Subscriber`?
Does this simply mean the subscriber's name?

event/event.go
```diff
@@ -0,0 +1,11 @@
+package event
+
+type Event interface {
+	GetEventType() string
+
+	GetEventAttributeKey() string
+
+	GetEventAttributeValue() string
+
+	GetEventHandler()
```
gyuguen:
```suggestion
	GetEventHandler() func() error
```

event/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+	"context"
+	"fmt"
+	"github.com/medibloc/panacea-doracle/config"
+	log "github.com/sirupsen/logrus"
+	rpchttp "github.com/tendermint/tendermint/rpc/client/http"
+	"time"
+)
+
+type PanaceaSubscriber struct {
+	WSAddr     string
```
gyuguen: I think this is an unnecessary field.
There seems to be no place to use this.

event/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+const (
+	RegisterOracle PanaceaEventStatus = 1
+)
+
+func (s *PanaceaSubscriber) Run(event ...PanaceaEventStatus) error {
+	log.Infof("start panacea event subscriber")
+	client, err := rpchttp.New(s.WSAddr, "/websocket")
+	if err != nil {
+		return err
+	}
```
gyuguen:
```suggestion
	
```

I think it is unnecessary.

server/server.go
```diff
@@ -49,6 +50,12 @@ func Run(conf *config.Config) error {
 	case <-signalCh:
 		log.Info("signal detected")
 	}
+
+	err = subscriber.Close()
+	if err != nil {
+		return err
+	}
```
gyuguen:
```suggestion
	if err := subscriber.Close(); err != nil {
		return err
	}
```

gyuguen (commented):
Thank you.
i left a few comment :)

event/register_oracle_event.go
```diff
@@ -0,0 +1,26 @@
...
+	return e.EventAttributeKey
+}
+
+func (e RegisterOracleEvent) GetEventAttributeValue() string {
+	return e.EventAttributeValue
+}
+
+func (e RegisterOracleEvent) GetEventHandler() {
+	panic("implement me")
+}
```
gyuguen:
```suggestion
func (e RegisterOracleEvent) GetEventHandler() func() error {
	return e.EventHandler
}
```

event/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+	Subscriber string
+	Client     *rpchttp.HTTP
+}
+
+// NewSubscriber generates a rpc http client with websocket address.
+func NewSubscriber(conf *config.Config) (*PanaceaSubscriber, error) {
+	client, err := rpchttp.New(conf.Panacea.WSAddr, "/websocket")
+	if err != nil {
+		return nil, err
+	}
```
0xHansLee: RPC client is also used in light client. I think we can share the RPC client for both light client and subscriber.
But it's out of scope of this PR, so let's refactor these when merging everything.

event/subscriber.go
0xHansLee: You can use `s.Client`

config/config.go
inchori: Yes, this is just a subscriber's name. It is used in `client.Subscribe(ctx, "subscriber's name", query)`.
How about delete this and leave a name empty?

event/subscriber.go
```diff
@@ -0,0 +1,99 @@
...
+	}, nil
+}
+
+type PanaceaEventStatus int32
+
+const (
+	RegisterOracle PanaceaEventStatus = 1
+)
+
+func (s *PanaceaSubscriber) Run(event ...PanaceaEventStatus) error {
```
0xHansLee: Do you have any specific reason to use `PanaceaEventStatus` instead of `Event` interface itself?

server/server.go
```diff
@@ -49,6 +50,11 @@ func Run(conf *config.Config) error {
 	case <-signalCh:
 		log.Info("signal detected")
 	}
+
+	if err = subscriber.Close(); err != nil {
+		return err
+	}
+
```
0xHansLee: If subscriber is needed only for close, you can close the subscriber in `cmd/start.go` using defer function.

config/config.go
youngjoon-lee: > How about delete this and leave a name empty?

+1

event/register_oracle_event.go
```diff
@@ -0,0 +1,26 @@
...
+	return e.EventAttributeKey
+}
+
+func (e RegisterOracleEvent) GetEventAttributeValue() string {
+	return e.EventAttributeValue
+}
+
+func (e RegisterOracleEvent) GetEventHandler() func() error {
+	return e.EventHandler
+}
```
youngjoon-lee:
```suggestion
type RegisterOracleEvent struct {}

func (e RegisterOracleEvent) GetEventType() string {
	return "message"
}

func (e RegisterOracleEvent) GetEventAttributeKey() string {
	return "action"
}

func (e RegisterOracleEvent) GetEventAttributeValue() string {
	return "RegisterOracle"
}

func (e RegisterOracleEvent) EventHandler() error {
	// TODO: Executing Voting Tx
	fmt.Println("RegisterOracle Event Handler")
	return nil
}
```
It's good to have the `Event` interface as you did.
You can implement this `RegisterOracleEvent` struct statically because the logic of these 4 functions won't be changed dynamically.

event/subscriber.go
inchori: I didn't review light client yet. When merging everything, it will be refactored! Thanks!

event/subscriber.go
inchori: The `PanaceaEventStatus` is used just for using enumerated events.

youngjoon-lee: I think the same

event/subscriber.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+	defer cancel()
+
+	for _, e := range event {
+		convertedEvent := convertEventStatusToEvent(e)
+		query := convertedEvent.GetEventType() + "." + convertedEvent.GetEventAttributeKey() + "=" + convertedEvent.GetEventAttributeValue()
+		txs, err := s.Client.Subscribe(ctx, s.Subscriber, query)
+		if err != nil {
+			return err
+		}
```
youngjoon-lee: It's better to create a context (with timeout) right before calling `s.Client.Subscribe`, if you intended to set the 1s timeout for sending a ''subscribe' request to the Panacea.
If you create a context outside of the loop and reuse it for 'subscribe' calls, the timeout error will occur unless all 'subscribe' calls are done in 1s.

event/register_oracle_event.go
inchori: Oh, I got it. It makes sense. Thank you!

event/subscriber.go
inchori: When I changed like following code, there has a warning in `defer.cancel()`.
The warning said `Possible resource leak, 'defer' is called in the 'for' loop`. Is it okay to change like following code?
```go
for _, e := range event {
		convertedEvent := convertEventStatusToEvent(e)
		query := convertedEvent.GetEventType() + "." + convertedEvent.GetEventAttributeKey() + "=" + convertedEvent.GetEventAttributeValue()

		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
		defer cancel()

		txs, err := s.Client.Subscribe(ctx, "", query)
		if err != nil {
			return err
		}
```

config/config.go
inchori: I just leave a subscriber's name field empty.

event/event.go
```diff
@@ -0,0 +1,11 @@
...
+
+type Event interface {
+	GetEventType() string
+
+	GetEventAttributeKey() string
+
+	GetEventAttributeValue() string
+
+	EventHandler() error
+}
```
gyuguen: What if we could add a `ResultEvent` to the handler to use?

```suggestion
package event

import ctypes "github.com/tendermint/tendermint/rpc/core/types"

type Event interface {
	GetEventType() string

	GetEventAttributeKey() string

	GetEventAttributeValue() string

	EventHandler(ctypes.ResultEvent) error
}
```

inchori: Yes, I agree.
So we can query or verify with `ResultEvent`.

inchori: Fixed in [133371d](https://github.com/medibloc/panacea-doracle/pull/15/commits/133371d7ed9d36e9098f466b94346c31aae260ce).

event/subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+	log.Infof("start panacea event subscriber")
+
+	for _, e := range events {
+		query := e.GetEventType() + "." + e.GetEventAttributeKey() + "=" + e.GetEventAttributeValue()
+
+		ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer cancel()
+
+		fmt.Println(query)
+
```
0xHansLee:
```suggestion
```

event/subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+		txs, err := s.Client.Subscribe(ctx, "", query)
+		if err != nil {
+			return err
+		}
+
+		e := e
```
0xHansLee:
```suggestion
```

event/subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+		go func() {
+			for range txs {
+				t := <-txs
+				fmt.Println("got ", t.Events)
+				_ = e.EventHandler(t)
+			}
```
0xHansLee:
```suggestion
			for t := range txs {
				fmt.Println("got ", t.Events)
				_ = e.EventHandler(t)
			}
```
You can use like this :)

event/subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+//func convertEventTypeToEvent(eventType string) Event {
+//	switch eventType {
+//	case types.EventTypeRegistrationVote:
+//		return RegisterOracleEvent{}
+//	default:
+//		return nil
+//	}
+//}
```
0xHansLee: you can delete this commented code block :)

event/subscriber.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (s *PanaceaSubscriber) Close() error {
+	log.Infof("closing panacea event subscriber")
+	if err := s.Client.Stop(); err != nil {
+		return err
+	}
+	return nil
```
0xHansLee:
```suggestion
	return s.Client.Stop()
```
If you don't wrap the error from `s.Client.Stop()` or this `Close()` only call the `s.Client.Stop()`, you can just return it.

inchori: Thanks for comment!

event/subscriber.go
inchori: If I deleted this, there has a lint issue. So I'll check and fix it and ping re-review to you.

inchori: Fixed in [37ab159](https://github.com/medibloc/panacea-doracle/pull/15/commits/37ab159f1a1202f0347b14dedb4374f52e42c5fa).

0xHansLee (approved):
lgtm

gyuguen (approved):
LGTM 👍

event/subscriber.go
youngjoon-lee: Yeah. The warning matters. Let me check how the `s.Client.Subscribe` is implemented. 1 sec.

youngjoon-lee: Alright. I just checked how the internal of `Subscribe(ctx, ...)` works. The `ctx` is used only when sending a packet which initializes the subscription. After the subscription is established, the `ctx` is not used. It means that the subscriber will subscribe new events endlessly as we intend. All good.

Nevertheless, it's important to set a timeout using `ctx` as you did, so that we can get a timeout error if it takes a long time to send a initialization packet.
To avoid the warning in terms of `defer`, I would recommend you to define a new function like:
```go
func subsribe(query string) {
          ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
          defer cancel()

          txs, err := s.Client.Subscribe(ctx, "", query)
          if err != nil {
              return err
          }

          go func(e Event) {
              for t := range txs {
                  fmt.Println("got ", t.Events)
                  _ = e.EventHandler(t)
              }
          }(e)
}
```

Then, you can call this function in your for loop.

youngjoon-lee: Actually, you don't need to set timeouts for now when communicating with Tendermint. There are lots of logics in our doracle which communicate with Tendermint, but we aren't setting any timeout for now. Later, we have to define a new parameter `panacea_write_timeout` in our `config.toml`, and use it everywhere. But, let's go with your code for now.

inchori: Okay I got it!. Thank you.
I refactored in [67719da](https://github.com/medibloc/panacea-doracle/pull/15/commits/67719daf9bf6e90041cceae9d5adeb4c60315bc8).

service/service.go
```diff
@@ -0,0 +1,64 @@
...
+	OracleAccount *panacea.OracleAccount
+	OraclePrivKey *btcec.PrivateKey
+	UniqueID      string
+
+	QueryClient *panacea.QueryClient
+	GrpcClient  *panacea.GrpcClient
+}
+
+func New(conf *config.Config) (*Service, error) {
+	homeDir, err := os.UserHomeDir()
```
gyuguen, line 27: When I tested, the homeDir set here and [another homeDir](https://github.com/medibloc/panacea-doracle/blob/main/cmd/doracled/cmd/root.go#L36) are different.

```
# gen-oracle-key command
/data/.panacea/oracle_priv_key.sealed

# run command
/data/oracle_priv_key.sealed
```

event/subscriber.go
```diff
@@ -37,23 +37,32 @@ func (s *PanaceaSubscriber) Run(events ...Event) error {
 	for _, e := range events {
 		query := e.GetEventType() + "." + e.GetEventAttributeKey() + "=" + e.GetEventAttributeValue()
```
youngjoon-lee: I think we can move this line 38 into the `subscribe()` function that you just added.

youngjoon-lee (approved):
lgtm

service/service.go
0xHansLee: I'll check and fix it. Thank you gyuguen.

gyuguen: Good.

It would be good to commonize the part that gets thest paths later.

0xHansLee: Agreed. I'll do that in another PR.

gyuguen: I'm making a function that can set homeDir (`doracle start --home /data/.doracle2`), so let's do this together. :)

PR #16, opened 2022-08-25
feat: implementation of get-oracle-key cmd

### LINKED ISSUE #14: Implement a `get-oracle-key` cmd
After oracle is registered successfully, oracle can retrieve the oracle private key securely.
The detailed steps are listed as below:

- get `EncryptedOraclePrivKey` from Panacea
- decrypt the `EncryptedOraclePrivKey` using node private key to retrieve oracle private key
- seal the decrypted oracle private key
- store the sealed oracle private key

0xHansLee:
Close #14 

When voting for registration of oracle passed, the oracle can be shared the oracle private key using this cmd.

This cmd do the followings:
- check if `node_priv_key.sealed` file exists
- get node private key from the sealed file
- compare the node key is equal to the one from the vote (oracle registration in Panacea)
- check the oracle registration vote status
  - if the vote is passed, get the `EncryptedOraclePrivKey` and decrypt, seal, and store it as `oracle_priv_key.sealed` file.
  - If not, returns error. we can improve this cmd when voting is in progress by adding retry.


gyuguen (approved):
LGTM 👍

audtlr24 (approved):
LGTM!

PR #19, opened 2022-08-29
feat: add service to `start` cmd

### LINKED ISSUE #17: skeleton code for `start` cmd

0xHansLee:
added service to `start` cmd, which includes the below fields:
- configuration
- oracle account info (priv + pub key) for signing Tx
- unique ID to be used to verify remote report
- query client for querying data and verifying it
- gRPC client for broadcasting Tx

The service can be used in event handler for data verification and signing transactions.


cmd/doracled/cmd/root.go
```diff
@@ -30,7 +49,7 @@ func init() {
 	rootCmd.PersistentFlags().StringVar(&homeDir, "home", defaultAppHomeDir, "application home directory")
 
 	rootCmd.AddCommand(initCmd)
-	rootCmd.AddCommand(startCmd)
+	rootCmd.AddCommand(startCmd())
 	rootCmd.AddCommand(genOracleKeyCmd)
 	rootCmd.AddCommand(verifyReport)
 	rootCmd.AddCommand(RegisterOracleCmd())
```
gyuguen: There seems to be a mix of ways to create Cmd.
It would be good to merge them into one later (probably as a function?)

gyuguen (approved):
LGTM 👍

cmd/doracled/cmd/root.go
0xHansLee: Right. I'll remember and refactor it. Thanks!

inchori (approved):
LGTM

PR #21, opened 2022-08-30
feat: add refresh function to update light block

audtlr24:
Add `refresh` function to update light block, when the last light block has been updated more than `trustPeriod` *2 /3.
If we call `NewQueryClient`, the `refresh` is called every minute with go routine.

Another Change
- refactor `GetStoreData` that check whether or not to wait `BlockPeriod` (= 6 seconds) to verify merkle proof.


panacea/query_client.go
```diff
@@ -72,6 +73,14 @@ func NewQueryClient(ctx context.Context, chainID, rpcAddr string, trustedBlockHe
 	if err != nil {
 		return nil, err
 	}
+	// call refresh every minute
+	go func() {
+		for {
+			time.Sleep(1 * time.Minute)
+			err = refresh(ctx, lc, trustOptions.Period)
+			logrus.Info("light client refresh error: ", err)
```
gyuguen: I think that error log will occur even if no error occurs.
So how about modifying it like this:
```suggestion
			if err := refresh(ctx, lc, trustOptions.Period); err != nil {
			    logrus.Errorf("light client refresh error: %v", err)
			}
```

gyuguen: And we have to think about the behavior when refresh fails.
As for now, whether to output only the log or terminate the process.

audtlr24: Ah, I got it. Thanks! I reflect it.

panacea/query_client.go
```diff
@@ -112,6 +123,30 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
...
+	}
+	lastBlockTime := lastBlock.Time
+	currentTime := time.Now()
+	timeDiff := currentTime.Sub(lastBlockTime)
+	if timeDiff > trustPeriod*2/3 {
+		logrus.Info("update latest light block")
+		_, err = lc.Update(ctx, time.Now())
+		if err != nil {
+			return err
+		}
```
gyuguen: It would be good to change it like this.
```suggestion
		if _, err := lc.Update(ctx, time.Now()); err != nil {
			return err
		}
```

panacea/query_client.go
```diff
@@ -131,15 +166,26 @@ func (q QueryClient) GetStoreData(ctx context.Context, storeKey string, key []by
...
-	time.Sleep(blockPeriod) // AppHash for query is in the next block, so have to wait until the next block is confirmed
-
-	// wait a creation of the next block
-	nextTrustedBlock, err := q.LightClient.VerifyLightBlockAtHeight(ctx, trustedBlock.Height+1, time.Now())
+	// get nextTrustedBlock from LightClient Primary.
+	// It returns a new light block on a successful update. Otherwise, it returns nil
+	nextTrustedBlock, err := q.LightClient.Update(ctx, time.Now())
 	if err != nil {
 		return nil, err
 	}
```
gyuguen: If the next block does not exist, an error is occured and this process ends.
I think you need to make the code to retry a few times when an error that the block is not created occurs.

panacea/query_client.go
```diff
@@ -112,6 +123,30 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
 	}, nil
 }
 
+// refresh update light block, when the last light block has been updated more than trustPeriod * 2/3.
```
youngjoon-lee:
```suggestion
// refresh update light block, if the last light block has been updated more than trustPeriod * 2/3 ago.
```

panacea/query_client.go
youngjoon-lee: As gyuguen said, I also think it's better to retry the `VerifyLightBlockAtHeight` a few times, instead of calling the `q.LightClient.Update(...)` two times in this function. We're calling the `q.LightClient.Update(...)` at the beginning of this function. So, the `q.LightClient.Update(...)` at line 171 will return `nil` with a very high probability. Then, we're gonna sleep 6 secs that decreases performance.

So, I would suggest something like:
```go
for {
  nextTrustedBlock, err = q.LightClient.VerifyLightBlockAtHeight(ctx, trustedBlock.Height+1, time.Now())
  if errors.Is(err, provider.ErrlightBlockNotFound) { // TODO: needs to be tested
    time.Sleep(1 * time.Second)
    continue
  } else if err != nil {
    return nil, err
  }
  break
}
```

panacea/query_client.go
```diff
@@ -112,6 +123,30 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
...
+		return err
+	}
+	lastBlock, err := lc.TrustedLightBlock(lastBlockHeight)
+	if err != nil {
+		return err
+	}
+	lastBlockTime := lastBlock.Time
+	currentTime := time.Now()
+	timeDiff := currentTime.Sub(lastBlockTime)
+	if timeDiff > trustPeriod*2/3 {
```
0xHansLee, line 158: Should this `timeDiff` be less than `trustedPeriod`?
If the `timeDiff` is greater than `trustedPeriod`, we need to handle differently (maybe skipping verification?)

panacea/query_client.go
```diff
@@ -112,6 +123,30 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
...
+		logrus.Info("update latest light block")
```
0xHansLee: We usually alias this `logrus` to `log`

audtlr24: Thanks, I fixed it in [2235390](https://github.com/medibloc/panacea-doracle/pull/21/commits/22353900e45ca233fe82c98c1cc666f27fc3b703).

panacea/query_client.go
audtlr24: Thanks, I fixed it in [2235390](https://github.com/medibloc/panacea-doracle/pull/21/commits/22353900e45ca233fe82c98c1cc666f27fc3b703).

panacea/query_client.go
audtlr24: @youngjoon-lee @gyuguen Thanks, I fixed it in [07ede15](https://github.com/medibloc/panacea-doracle/pull/21/commits/07ede15f1487ce888f1e69000e6dc3c26eda21ec). Thank you for reviewing again.

audtlr24:
@youngjoon-lee @gyuguen @H4NLee I found a concurrency problem in the part running the light client. When multiple queries are executed at the same time like `TestMultiGetAddress` , I checked the problem in light client `verify` where the latest block is fetched. 

So, I tried to solve this problem through mutex. The problem has been resolved, and please review again. Thanks!

panacea/query_client.go
audtlr24: Yes, you're right. If `timeDiff` is greater than `trustedPeriod`, it returns error when `Update` (exactly,  expired error occurs in the `verify`). 
It looks like we can solve it, but I think the logic is complicated, and I'm not sure if that logic is needed yet. 
Let's talk together.

panacea/query_client.go
```diff
@@ -13,13 +14,16 @@ import (
 	aoltypes "github.com/medibloc/panacea-core/v2/x/aol/types"
 	"github.com/medibloc/panacea-doracle/config"
 	sgxdb "github.com/medibloc/panacea-doracle/store/sgxleveldb"
-	"github.com/tendermint/tendermint/libs/log"
+	log "github.com/sirupsen/logrus"
+	tmlog "github.com/tendermint/tendermint/libs/log"
+	"github.com/tendermint/tendermint/libs/sync"
```
youngjoon-lee: I recommend you to use the `"sync"` package provided by Go by default.
https://pkg.go.dev/sync
https://go.dev/tour/concurrency/9

Actually, the `"github.com/tendermint/tendermint/libs/sync"` is a wrapper of the `"sync"`, but I think you don't have any particular reason to use the Tendermint one.

panacea/query_client.go
```diff
@@ -98,46 +103,108 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
...
 // GetStoreData get data from panacea with storeKey and key, then verify queried data with light client and merkle proof.
 // the returned data type is ResponseQuery.value ([]byte), so recommend to convert to expected type
 func (q QueryClient) GetStoreData(ctx context.Context, storeKey string, key []byte) ([]byte, error) {
+	var queryHeight int64
+
+	// get recent light block
+	// if the latest block has already been updated, get LastTrustedHeight
+	q.mutex.Lock()
 	trustedBlock, err := q.LightClient.Update(ctx, time.Now())
+	q.mutex.Unlock()
```
youngjoon-lee: The best practice of unlocking the mutex is using `defer q.mutex.Unlock()`, so that the mutex can be unlocked in any cases regardless what happened in this function.

You may think that it's safe to unlock it right after calling the `q.LightClient.Update(...)` (aka. [critical section](https://en.wikipedia.org/wiki/Critical_section)). In most cases, yes, you're right. But, the critical section can raise a [panic](https://go.dev/blog/defer-panic-and-recover). Then, this `GetStoreData()` function will be returned immediately, and the panic will be thrown to the caller of `GetStoreData()`. If so, the `q.mutex.Unlock()` won't be executed because the process won't reach to the line 165. 
In our current codes, your implementation could be safe because we don't [recover](https://go.dev/blog/defer-panic-and-recover) panics anywhere. But, we can probably do it in the future. Who knows. 

So, it's much better to use `defer` to unlock mutexes safely. The `defer` statement is executed even if a panic occurs. You can see my example: https://gist.github.com/youngjoon-lee/8c3304b556ac6ce6a3ef856a91950f16.

Because the `GetStoreData()` function is big, I would recommend you to define another function which just calls `q.LightClient.Update()` with mutex.
```go
func (q QueryClient) safeUpdateLightClient(...) {
  q.mutex.Lock()
  defer q.mutex.Unlock()

  return q.LightClient.Update(...)
}
```

You can use this practice for the `q.LightClient.VerifyLightBlockAtHeight` at line 195 as well.

panacea/query_client.go
audtlr24: Ah, I got it. It's my mistake. Thanks! I fixed it in [0a8904b](https://github.com/medibloc/panacea-doracle/pull/21/commits/0a8904ba57ea69fd3fd4694cc5779fa85f47d8f3).

panacea/query_client.go
audtlr24: Thanks! I fixed it in [0a8904b](https://github.com/medibloc/panacea-doracle/pull/21/commits/0a8904ba57ea69fd3fd4694cc5779fa85f47d8f3). I did not considered panic, thanks for let me know.

youngjoon-lee (approved):
lgtm!

0xHansLee (approved):
LGTM 👍

PR #23, opened 2022-08-30
feat: store sealed trusted block info for light client initialization

0xHansLee:
For light client initialization, the trusted block info was given in `register-oracle` cmd by a user (operator).
After the vote for oracle registration is passed (other oracles also check if the given trusted block info is valid), the trusted block info should not be changed.
To do that, the trusted block info is sealed and stored in a file named `trusted_block_info.sealed`.

- `register-oracle` : seal and store `trusted_block_info.sealed`
- `get-oracle-key` : unseal from `trusted_block_info.sealed` and utilize the trusted block info to initialize light client


cmd/doracled/cmd/get_oracle_key.go
```diff
@@ -16,69 +18,96 @@ import (
...
+			oracleAccount, err := getOracleAccount(cmd, conf.OracleMnemonic)
+			if err != nil {
+				return fmt.Errorf("failed to get oracle account from mnemonic: %w", err)
+			}
+
+			trustedBlockInfo, err := readTrustedBlockInfoFile()
+			if err != nil {
+				return fmt.Errorf("failed to read trusted block info file: %w", err)
+			}
+			fmt.Printf("%v will be used in query client", trustedBlockInfo)
```
gyuguen, line 67: There doesn't seem to be anywhere that uses trusted BlockInfo, where do we use it?

gyuguen: Oh, It will use in the future. :)

gyuguen:
As we discussed, it is judged that the part that seals the trusted block info is not necessary. :)

0xHansLee:
Close this PR because there is no need to seal and store the trusted block info.
We decided to use `ligth.NewClientFromTrustedStore()` to load the existing light client after initialization

PR #24, opened 2022-08-30
chore: use tx response raw log instead of logs

0xHansLee:
When Tx fails, there is no info in `TxResponse.Logs`. Instead, changed to `TxResponse.RawLog` which has detailed info of transaction.

FYI) I also increased the default gas limit and fee because with the previous values, tx fails with an error of `out of gas`.


inchori (approved):
LGTM Thanks! 👍

gyuguen (approved):
LGTM 👍

youngjoon-lee (approved):

PR #25, opened 2022-08-31
build: add artifacts for production deployments

youngjoon-lee:
From my tests with EGo v1.0.1, the `exe` path in the `enclave.json` is important only when running the `ego sign ...`. After that, it seems that we can move the binary to anywhere we want.


gyuguen (approved):
LGTM 👍

0xHansLee (approved):
👍

audtlr24 (approved):
lgtm! Thanks.

PR #26, opened 2022-09-01
refactor: change uniqueID encoding from base64 to hex

0xHansLee:
Refactoring of changing encoding of `uniqueID` from base64 to hex

Ego uses hex encoding for uniqueID. On the other hand, we used base64 encoding uniqueID.
Thus, it was difficult to use the hex-encoded uniqueID, which can be obtained from the `ego uniqueid <executable>` cmd, for interaction with Panacea because different encoded uniqueID was used to store data (e.g. oracle registration).

By using the same encoding scheme, user can utilize the encoded uniqueID obtained from `ego uniqueid` cmd as it is.

### DIFF
--- a/cmd/doracled/cmd/get_oracle_key.go
+++ b/cmd/doracled/cmd/get_oracle_key.go
@@ -2,7 +2,7 @@ package cmd
 
 import (
 	"bytes"
-	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


inchori (approved):
lgtm

audtlr24 (approved):
lgtm!

gyuguen (approved):
LGTM 👍

PR #27, opened 2022-09-01
feat: added to use homeDir path with home flag

gyuguen:
* Improved to use the path of home in flags
* Added common function to get path information


0xHansLee:
Why don't we make a context (that includes homeDir) as cosmos-sdk did?

gyuguen:
> Why don't we make a context (that includes homeDir) as cosmos-sdk did?

This is because the path to save settings and data information is set differently.
For example, someone may want to store it in `/data/.doracle`, and someone may want to store it in `/data/.doracle_product`.
In other words, to run multiple `panacea-doracle` nodes on one device.

When we test, we will be able to run multiple `panacea-doracle`s on one machine.

gyuguen:
Actually we are already getting the --home value as a global variable.
`doracle --h`

image

gyuguen:
Oh I misunderstood.

gyuguen:
It seems that we are not using context like cosmos.
What does @H4NLee  mean by context?

gyuguen:
Umm.. like this?
```go
clientCtx, err := client.GetClientQueryContext(cmd)
homeDir := clientCtx.HomeDir
```

0xHansLee:
Yes! Just wanna make it reusable the code block and refer the cosmos-sdk. wdyt?
```go
homeDir, err := cmd.Flags().GetString(flags.FlagHome)
		if err != nil {
			return err
		}
```

gyuguen:
> Yes! Just wanna make it reusable the code block and refer the cosmos-sdk. wdyt?
> 
> ```go
> homeDir, err := cmd.Flags().GetString(flags.FlagHome)
> 		if err != nil {
> 			return err
> 		}
> ```
Thanks for your comment.
I improved it to use context. :)

youngjoon-lee:
In general, looks good to use the context, but I will revisit this PR after CI (lint) passes.

youngjoon-lee (changes requested):
I just read the entire codebase, but I've seen there are many configuration-related things scattered everywhere (especially in the light client implementation). Before we merge this PR, I think we should refactor them.

I'll introduce a new PR soon.

youngjoon-lee:
> I'll introduce a new PR soon.

I just opened a PR #36  which improve this PR.

When refactoring codes, I realized that `client.Context` makes things complicated for now because we don't have any specific purpose of using `client.Context`. That could be used for unit testing in the future, but for now, let's keep it simple. I wrote detailed reasons into the PR #36.

Please review and merge it into this PR if you like it.

youngjoon-lee:
> Please review and merge it into this PR if you like it.

I just merged the PR #36 to this PR. @gyuguen

0xHansLee (approved):
lgtm

gyuguen:
> > Please review and merge it into this PR if you like it.
> 
> I just merged the PR #36 to this PR. @gyuguen

Thx!

youngjoon-lee (approved):
thank you

PR #28, opened 2022-09-01
feat: added tx success log

gyuguen:
I'm adding the necessary logs because it's hard to tracking tx.


0xHansLee (approved):
You're right. Thank you!

audtlr24 (approved):
Thanks!

youngjoon-lee (approved):

PR #29, opened 2022-09-05
test: sgx-related unit tests

0xHansLee:
I added two test-cases:
- failure case : verification of remote report with different data
- success case : `SealToFile` and `UnsealFromFile`

### DIFF
--- a/sgx/seal_test.go
+++ b/sgx/seal_test.go
@@ -11,11 +11,12 @@ import (
 func TestSealAndUnsealFile(t *testing.T) {
 	data := []byte("hello world")
 
-	dir, err := os.MkdirTemp("", "test")
+	testDir := "/test"
+	err := os.MkdirAll(testDir, os.ModePerm)
 	require.NoError(t, err)
-	defer os.RemoveAll(dir)
+	defer os.RemoveAll(testDir)
 
-	dataPath := filepath.Join(dir, "temp.sealed")
+	dataPath := filepath.Join(testDir, "temp.sealed")
 
 	err = sgx.SealToFile(data, dataPath)
 	require.NoError(t, err)
--- a/sgx/seal_test.go
+++ b/sgx/seal_test.go
@@ -1,0 +1,27 @@
+package sgx_test
+
+import (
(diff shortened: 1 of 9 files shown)


sgx/seal_test.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/stretchr/testify/require"
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSealAndUnsealFile(t *testing.T) {
+	data := []byte("hello world")
+
+	testDir := "/test"
```
0xHansLee, line 14: I used a temp dir, "/test", for testing.
Since `ego sign <executable>`, not signing using `enclave.json`, the default `enclave.json` is created.
So `$HOME` env var is not set in test environment and I can't use home dir.
If you have any better idea, please let me know 🙏

youngjoon-lee: It's okay to use the `/test` because that directory doesn't need to be permanently kept after this test is finished. 

In general, you can create a temporary file with a random filename. In Linux, temporary files are created in the `/tmp` (aka. tmpfs). 
https://gist.github.com/youngjoon-lee/c1d79ddc83e993b3fba91631b034083a

youngjoon-lee (approved):

gyuguen (approved):
LGTM 👍

sgx/seal_test.go
0xHansLee: Right. I first tried to use `os.MkdirTemp` which makes directory under `/tmp`.
But I got an error of `stat /tmp: no such file or directory`. Thus I should have created directory `/tmp` as well.
It seems that there is no directory of `/tmp` in enclave.
If it would be better to change the directory name to `/tmp` instead of `/test`, please let me know 😄

gyuguen: According to this [link](https://docs.edgeless.systems/ego/#/reference/config?id=mounts), the default enclave.json uses `memfs` type and only empty `/` exists.

youngjoon-lee: Ja. Then, keep your code as it is, because this test won't work on non-SGX environments anyway.

PR #30, opened 2022-09-05
feat: add LoadQueryClient function

audtlr24:
Add `LoadQueryClient` function that can get `queryClient` from trustedStore.

Since the light block that has already been saved is used (by `NewQueryClient`), there is no need to use trusted block info as a parameter.

### DIFF
--- a/panacea/query_client.go
+++ b/panacea/query_client.go
@@ -67,7 +67,7 @@ func newQueryClient(ctx context.Context, config *config.Config, info *TrustedBlo
 		return nil, err
 	}
 
-	pv, err := tmhttp.New(chainID, config.Panacea.PrimaryAddr)
+	pv, err := tmhttp.New(chainID, config.Panacea.LightClientPrimaryAddr)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


panacea/query_client.go
```diff
@@ -112,7 +113,68 @@ func NewQueryClient(ctx context.Context, config *config.Config, info TrustedBloc
...
+	return &QueryClient{
+		RpcClient:         rpcClient,
+		LightClient:       lc,
+		interfaceRegistry: makeInterfaceRegistry(),
+		sgxLevelDB:        db,
+		mutex:             &lcMutex,
+	}, nil
+}
+
+func LoadQueryClient(ctx context.Context, config *config.Config) (*QueryClient, error) {
```
gyuguen, line 55: Among these, there is duplicate code that is quite similar to NewQueryClient.
I'll see if there's a way to improve

gyuguen: @audtlr24 I made the improved code PR. #39 
If you think it has been improved, please approve it.  :)

gyuguen (approved):
LGTM 👍

0xHansLee (approved):
lgtm
It seems that the conflict needs to be resolved :)

PR #31, opened 2022-09-05
refactor: reorganize panacea configs

youngjoon-lee:
- We don't need to have `websocket-addr` and `rpc-addr` together. Both are the same.
- I renamed the `witnesses-addr` to `witness-addrs`.
- I put a `light-client-` prefix to the `primary-addr` and `witness-addrs`.
  - with parsing `witness-addrs` immediately when unmarshalling the toml file.
- Also, I don't recommend to expose `*.gopanacea.org` addrs in our codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,8 +167,8 @@ For that, the public key and its remote report are required.
 
 ```json
 {
-  "public_key_base_64" : "<base64-encoded-public-key>",
-  "remote_report_base_64": "<base64-encoded-remote-report>"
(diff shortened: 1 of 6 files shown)


0xHansLee (approved):
👍

gyuguen (approved):
LGTM 👍

inchori (approved):
LGTM 👍

audtlr24 (approved):
lgtm!

PR #32, opened 2022-09-05
feat: verify remote report for oracle registration event

### LINKED ISSUE #18: Implement a remote report verifier when detecting register oracle event
- When detecting `RegisterOracle` event, the subscriber needs to implement a remote attestation from oracle who want to register.
- Verifying a remote report will be implemented in `register_oracle_event.go` and executed in `EventHandler()`.

inchori:
- Verify Remote Report for Oracle Registration
- Get `nodePubKey` and `nodePubKeyRemoteReport` from querying `panacea-core`
- If the verification is passed, Vote Tx will be executed. (It will be implemented next branch)

### DIFF
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -28,7 +28,7 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent, svc *service
 	addressValue := event.Events[types.EventTypeRegistrationVote+"."+types.AttributeKeyOracleAddress][0]
 	fmt.Println(addressValue)
 
-	oracleRegistration, err := svc.GrpcClient.GetOracleRegistration(svc.UniqueID, addressValue)
+	oracleRegistration, err := svc.GrpcClient.GetOracleRegistration(addressValue, svc.UniqueID)
 	if err != nil {
 		return err
 	}
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -23,6 +23,11 @@ func (e RegisterOracleEvent) GetEventAttributeValue() string {
 
 func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
 	// TODO: Verifying Remote Attestation and Executing Vote Txs
+	for _, e := range event.Events {
+		for _, k := range e {
+			fmt.Println(k)
+		}
+	}
 
 	fmt.Println("RegisterOracle Event Handler")
 	return nil
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -32,19 +32,11 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent, svc *service
 		return err
 	}
 
-	err = verifyRemoteReport(oracleRegistration, svc.EnclaveInfo)
+	err = sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, oracleRegistration.NodePubKey, *svc.EnclaveInfo)
 	if err != nil {
 		return err
 	}
 
 	// TODO: Executing Vote Txs
-	return nil
-}
-
-func verifyRemoteReport(oracleRegistration *types.OracleRegistration, selfEnclaveInfo *sgx.EnclaveInfo) error {
-	err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, oracleRegistration.NodePubKey, *selfEnclaveInfo)
-	if err != nil {
-		return err
-	}
 	return nil
 }
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -1,6 +1,7 @@
 package event
 
 import (
+	"encoding/hex"
 	"github.com/medibloc/panacea-core/v2/x/oracle/types"
 	"github.com/medibloc/panacea-doracle/service"
 	"github.com/medibloc/panacea-doracle/sgx"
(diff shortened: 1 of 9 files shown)


event/register_oracle_event.go
```diff
@@ -21,9 +23,32 @@ func (e RegisterOracleEvent) GetEventAttributeValue() string {
...
+
+	// TODO: Executing Vote Txs
+	return nil
+}
+
+func verifyRemoteReport(oracleRegistration *types.OracleRegistration) error {
+	selfEnclaveInfo, err := sgx.GetSelfEnclaveInfo()
+	if err != nil {
+		return err
+	}
```
0xHansLee: You don't need to get self enclave info here.
Actually, `service.Service`'s `UniqueID` is from the self enclave info ([link](https://github.com/medibloc/panacea-doracle/blob/f4b0740670764a08bc97a85dc9d76ce66b154048/service/service.go#L39)).
So you can refactor the `service.Service` struct to have the whole enclave info, and utilize it.

event/subscriber.go
```diff
@@ -57,7 +57,7 @@ func (s *PanaceaSubscriber) subscribe(event Event) error {
 
 	go func(event Event) {
 		for tx := range txs {
-			if err := event.EventHandler(tx); err != nil {
+			if err := event.EventHandler(tx, s.Service); err != nil {
```
0xHansLee: I think we can have better structure for service, subscriber, and event. 
Maybe put subscriber into the service as @youngjoon-lee said?
I'll also think about it.

youngjoon-lee: Since inchori is trying to using polymorphism by using the `Event` interface, I think there is no other way than to pass the `service.Service` to the `event.EventHandler(...)`. Or, you can also define each `*Event` structs to contain a `service.Service` as I did in the [PoC](https://github.com/youngjoon-lee/doracle-poc/blob/1a7efda7b4d0248ae4362120d8a861a7ddcc273d/pkg/dhub/event/join.go#L19). I'm okay with both ways.

But still, as @H4NLee said, it would be great to put a subscriber object into the `service.Service`.

youngjoon-lee: Ah then, it will cause the import cycle. Let me think.

youngjoon-lee: I just refactored this. 
https://github.com/medibloc/panacea-doracle/pull/37
Please review and merge it into your PR if you like it.

youngjoon-lee: I just closed the PR #37 after discussing with @gyuguen. I think we cannot refactor the `Service` nicely in the current milestone. Let's go with what inchori implemented for now. We can revisit this after designing the spec of the next milestone.

event/register_oracle_event.go
inchori: Fixed in [b127b4f](https://github.com/medibloc/panacea-doracle/pull/32/commits/b127b4f4df1c8c29eed5333bb4fe8d495a738cbb).

event/register_oracle_event.go
```diff
@@ -21,9 +24,27 @@ func (e RegisterOracleEvent) GetEventAttributeValue() string {
...
-	fmt.Println("RegisterOracle Event Handler")
+	oracleRegistration, err := svc.GrpcClient.GetOracleRegistration(addressValue, hex.EncodeToString(svc.EnclaveInfo.UniqueID))
+	if err != nil {
+		return err
+	}
+
+	err = verifyRemoteReport(oracleRegistration, svc.EnclaveInfo)
+	if err != nil {
+		return err
+	}
```
0xHansLee:
```suggestion
	if err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, oracleRegistration.NodePubKey, *svc.EnclaveInfo); err != nil {
		return err
	}
```

You can use the `sgx.VerifyRemoteReport` directly :)

inchori: Thanks. Clean code!
Fixed in [f4c315e](https://github.com/medibloc/panacea-doracle/pull/32/commits/f4c315e54e62b38170ac6d964dbddbea9c112c4d).

youngjoon-lee:
@inchori
Just merged my PR #37 to yours: [8ea22a5](https://github.com/medibloc/panacea-doracle/pull/32/commits/8ea22a5710427d8019ad0558acc65782b39cd00e). So, we are in the same boat :sunglasses:.

service/service.go
```diff
@@ -1,29 +1,32 @@
...
-	"os"
-	"path/filepath"
+	log "github.com/sirupsen/logrus"
 )
 
 type Service struct {
-	Conf          *config.Config
-	OracleAccount *panacea.OracleAccount
-	OraclePrivKey *btcec.PrivateKey
-	UniqueID      string
```
gyuguen, line 17: I think it would be better to leave the `UniqueID` alone. This is because there is an inconvenience of having to encode in hex every time to use this.

youngjoon-lee: Yeah. I think so, but also don’t wanna have a duplicate value along with enclaveInfo. 
Maybe we can define a `enclaveInfo.UniqueIDHex()`. I think the performance of the hex encoding is not a big issue. So, we can improve the convenience by defining that function.

gyuguen: Yep. That seems to be the better way.

event/register_oracle_event.go
```diff
@@ -22,8 +38,19 @@ func (e RegisterOracleEvent) GetEventAttributeValue() string {
...
-	// TODO: Verifying Remote Attestation and Executing Vote Txs
+	addressValue := event.Events[types.EventTypeRegistrationVote+"."+types.AttributeKeyOracleAddress][0]
+
+	uniqueID := hex.EncodeToString(e.reactor.EnclaveInfo().UniqueID)
+	oracleRegistration, err := e.reactor.GRPCClient().GetOracleRegistration(addressValue, uniqueID)
+	if err != nil {
+		return err
+	}
+
+	err = sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, oracleRegistration.NodePubKey, *e.reactor.EnclaveInfo())
```
gyuguen: When I tested it, it fails here.
```
time="2022-09-07T02:41:08Z" level=error msg="failed to handle event 'message.action='RegisterOracle'': invalid data in the report"
```

We hashed the NodePubKey to SHA256 and put into the remoteReport.
https://github.com/medibloc/panacea-doracle/blob/main/cmd/doracled/cmd/register_oracle.go#L167

inchori: Fixed in [4e92461](https://github.com/medibloc/panacea-doracle/pull/32/commits/4e92461808d4f2de1e9d67ce096d89cf9fb3a1c6).

0xHansLee (approved):
lgtm

gyuguen (approved):
LGTM 👍

PR #34, opened 2022-09-05
refactor: change `base_64` to `base64`

youngjoon-lee:
Please let me know if you think these base64 things should be changed to the hex format.


0xHansLee (approved):
I think we don't need to use hex encoding except `uniqueID`.

gyuguen (approved):
LGTM 👍 
I will also fit this spec to panacea-core.

PR #36, opened 2022-09-05
refactor: simplify the `--home` logic and make `Config` as a SSOT

youngjoon-lee:
Based on @gyuguen 's PR #27 

- I think we don't need a `client.Context`. The `cobra.Flags` contains everything that we need throughout the entire command tree. I understand why Cosmos SDK defined `client.Context` for unit testing, but I think the usage of `client.Context` is not intuitive. Let's keep codes simple for now, and let's adopt the `client.Context` after we set a proper unit testing strategy.
- I added configurable directory and file names into the `config.Config`.
   - Inspired by [Tendermint](https://github.com/tendermint/tendermint/blob/d7645628f13031746be27507d1673e132cf1f6e6/config/toml.go#L131)
   - Using this, I refactored the light client implementation (removing `init()`).


gyuguen (approved):
Thank you.
This has been very nicely simplified.
👍

0xHansLee (approved):
👍👍👍

audtlr24 (approved):
Thanks, lgtm!

PR #37, opened 2022-09-05
refactor: manage the lifecycle of `Subscriber` along with the `Service`

youngjoon-lee:
Based on @inchori 's PR #32 

- For easy graceful shutdown, let's manage the lifecycle of `Subscriber` along with the `Service`, as all other components.
  - To avoid import cycle errors, I introduces the `reactor` interface for each event type.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -55,26 +55,18 @@ func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, e
 		return nil, fmt.Errorf("failed to init subscriber: %w", err)
 	}
 
-	svc := &Service{
+	return &Service{
 		conf:          conf,
 		oracleAccount: oracleAccount,
 		oraclePrivKey: oraclePrivKey,
 		enclaveInfo:   selfEnclaveInfo,
 		grpcClient:    grpcClient.(*panacea.GrpcClient),
 		subscriber:    subscriber,
-	}
-
-	if err := svc.startSubscriptions(); err != nil {
-		return nil, fmt.Errorf("failed to start subscriptions: %w", err)
-	}
-
-	return svc, nil
+	}, nil
 }
 
-func (s *Service) startSubscriptions() error {
-	return s.subscriber.Run(
-		event.NewRegisterOracleEvent(s),
-	)
+func (s *Service) StartSubscriptions(events ...event.Event) error {
+	return s.subscriber.Run(events...)
 }
 
 func (s *Service) Close() error {
--- a/service/service.go
+++ b/service/service.go
@@ -3,27 +3,31 @@ package service
 import (
 	"encoding/hex"
 	"fmt"
+	"os"
+	"path/filepath"
+
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/medibloc/panacea-doracle/config"
 	"github.com/medibloc/panacea-doracle/crypto"
+	"github.com/medibloc/panacea-doracle/event"
 	"github.com/medibloc/panacea-doracle/panacea"
 	"github.com/medibloc/panacea-doracle/sgx"
 	"github.com/medibloc/panacea-doracle/types"
-	"os"
-	"path/filepath"
+	log "github.com/sirupsen/logrus"
 )
 
 type Service struct {
-	Conf          *config.Config
-	OracleAccount *panacea.OracleAccount
-	OraclePrivKey *btcec.PrivateKey
-	UniqueID      string
+	conf          *config.Config
+	oracleAccount *panacea.OracleAccount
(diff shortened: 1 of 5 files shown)


service/service.go
```diff
@@ -46,19 +50,50 @@ func New(conf *config.Config) (*Service, error) {
...
+
+	return svc, nil
 }
 
-func (s Service) Close() error {
+func (s *Service) startSubscriptions() error {
+	return s.subscriber.Run(
+		event.NewRegisterOracleEvent(s),
+	)
+}
```
gyuguen: My guess is that this command may need to add a number of different Events in addition to the RegistratorOracleEvent.
I think it would be good to pass the Event as a parameter.
Hm... we'll have to think a bit more about how to put it in this structure.

No problem at this milestone now. :)

gyuguen: If this service is only run with one command (eg `doracled start`) then you can put all the events here.

gyuguen (approved):
LGTM 👍

service/service.go
youngjoon-lee: I'm not sure if I understood your comment correctly.
The `service.go` is only for the `doracled start` which runs a background process. That is what the 'service' means.
And, the `subscriber.Run()` accepts multiple events as parameters. 

If we need to subscribe only one event in the different command such as `doracled hello..`, we can just create a `subscriber` separately without using the `Service`.

gyuguen: Yes, when I said `service` I meant `service.go`.
As you said, if `service.go` is only used for `doraced start` then there is no problem.

What I was thinking was a concern when this `service.go` was used in other commands as well.

youngjoon-lee: I got you. The name 'service' is quite confusing in this doracle because doracle doesn't provide any service to clients. It just subscribes events in background. And, as you said, doracle can be started with many different modes (or maybe not).
Hmm. then, we should change the structure. Let me think about it. I'm not sure if we can find nice structure now before starting the next milestone.

youngjoon-lee: Let's close this PR until we can know which new commands will be added and how they will use a subscriber and a grpc client. At that time, we can refactor this nicely.

youngjoon-lee:
Closing this PR after the discussion https://github.com/medibloc/panacea-doracle/pull/37/files#r963185976

youngjoon-lee:
Reopening since we can accept this small refactoring and improve this later at the next milestone.

service/service.go
```diff
@@ -46,19 +50,50 @@ func New(conf *config.Config) (*Service, error) {
...
+		oracleAccount: oracleAccount,
+		oraclePrivKey: oraclePrivKey,
+		uniqueID:      hex.EncodeToString(selfEnclaveInfo.UniqueID),
+		grpcClient:    grpcClient.(*panacea.GrpcClient),
+		subscriber:    subscriber,
+	}
+
+	if err := svc.startSubscriptions(); err != nil {
+		return nil, fmt.Errorf("failed to start subscriptions: %w", err)
+	}
```
0xHansLee: How about separating this `svc.startSubscriptions()`?
I think the `New` function just return a new svc. wdyt?

0xHansLee (approved):
I think these improvements are good for now.
So approved 👌

inchori (approved):
LGTM
I think this code organization is better than existing one. Also, the event handler's argument can be simplified for every events.

service/service.go
youngjoon-lee: Sorry. Didn't get what you mean. Could you elaborate it more?

youngjoon-lee: If you mean making the `startSubscriptions()` public, yes, we can. It would be good to make it public, so that callers can inject any event that they want. Also, to avoid unexpected behaviors. I will do it now.

0xHansLee: Yes. I meant make it public and start subscription outside of the `New` function.
maybe
```go
svc, err := service.New(conf)
if err != nil {
    return fmt.Errorf("failed to create service: %w", err)
}

if err := svc.StartSubscription(); err != nil {
    return err
}
```

I understood. Thanks!

youngjoon-lee: https://github.com/medibloc/panacea-doracle/pull/37/commits/1f864c0e441092bc87b837df22db46e931b6b597 Done

youngjoon-lee: I made the `service.StartSubscriptions()` as general as possible, but not sure if it would be helpful for next milestones. Let's see later. Maybe, we can make it better when the next specs are defined clearly.

youngjoon-lee:
Thank you guys. So, let's merge this PR. When we define specs of the next milestone clearly, we will probably be able to make this structure better.

gyuguen:
good!

PR #43, opened 2022-09-07
refactor: unexport private members in the `QueryClient`

youngjoon-lee:
Please try to not export members that are okay to be private.

### DIFF
--- a/panacea/query_client_test.go
+++ b/panacea/query_client_test.go
@@ -1,17 +1,17 @@
-package panacea_test
+package panacea
 
 import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"github.com/cosmos/cosmos-sdk/types/bech32"
-	"github.com/medibloc/panacea-doracle/config"
-	"github.com/medibloc/panacea-doracle/panacea"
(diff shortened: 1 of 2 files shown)


audtlr24 (approved):
Thanks, I will keep in mind!

panacea/query_client_test.go
```diff
@@ -1,17 +1,17 @@
-package panacea_test
+package panacea
```
0xHansLee, line 1: Is the package of this test `panacea`? not `panacea_test`?

youngjoon-lee: This should be `panacea` because some tests access private members in the `QueryClient` (actually, I made those members private in this PR).
If we really wanna test this package as an external user (aka. blackbox tests), we should use the `panacea_test` package. But for now, I think it's simpler to use `panacea`.

0xHansLee: I got it. Thank you!

0xHansLee (approved):

gyuguen (approved):
LGTM 👍

gyuguen:
This PR has been approved by 3 people, so I will merge it.

PR #44, opened 2022-09-07
feat: implement executing vote transaction after verify remote report

### LINKED ISSUE #33: Implement an Executing Vote Transaction for Oracle Registration
- When verifying for oracle registration passed, execute the vote transaction to `panacea-core` with `TxBuilder`.

inchori:
- Implement executing vote transaction after verify remote report.
- Executing Vote with `makeOracleRegistrationVote`, `generateTx`, and `broadcastTx` function in `event/register_oracle_event.go`.
- This feature works well when [this branch](https://github.com/medibloc/panacea-core/pull/400) merged into master.

### DIFF
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -1,9 +1,7 @@
 package event
 
 import (
-	"context"
 	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
@@ -31,6 +29,7 @@ type reactor interface {
 	OracleAcc() *panacea.OracleAccount
 	OraclePrivKey() *btcec.PrivateKey
 	Config() *config.Config
+	QueryClient() *panacea.QueryClient
 }
 
 func NewRegisterOracleEvent(r reactor) RegisterOracleEvent {
@@ -52,23 +51,13 @@ func (e RegisterOracleEvent) GetEventAttributeValue() string {
 func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
 	addressValue := event.Events[types.EventTypeRegistrationVote+"."+types.AttributeKeyOracleAddress][0]
 
-	uniqueID := hex.EncodeToString(e.reactor.EnclaveInfo().UniqueID)
-	oracleRegistration, err := e.reactor.GRPCClient().GetOracleRegistration(addressValue, uniqueID)
+	uniqueID := e.reactor.EnclaveInfo().UniqueIDHex()
+	oracleRegistration, err := e.reactor.QueryClient().GetOracleRegistration(addressValue, uniqueID)
 	if err != nil {
 		return err
 	}
 
-	trustedBlockInfo := panacea.TrustedBlockInfo{
-		TrustedBlockHeight: oracleRegistration.TrustedBlockHeight,
-		TrustedBlockHash:   oracleRegistration.TrustedBlockHash,
-	}
-
-	queryClient, err := panacea.NewQueryClient(context.Background(), e.reactor.Config(), trustedBlockInfo)
-	if err != nil {
-		return err
-	}
-
-	txBuilder := panacea.NewTxBuilder(*queryClient)
+	txBuilder := panacea.NewTxBuilder(*e.reactor.QueryClient())
 
 	voteOption := verifyReportAndGetVoteOption(oracleRegistration, e)
 
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -11,6 +11,7 @@ import (
 	"github.com/medibloc/panacea-doracle/config"
 	"github.com/medibloc/panacea-doracle/panacea"
 	"github.com/medibloc/panacea-doracle/sgx"
+	log "github.com/sirupsen/logrus"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 )
 
@@ -57,41 +58,33 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
 
 	nodePubKeyHash := sha256.Sum256(oracleRegistration.NodePubKey)
 
-	err = sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, nodePubKeyHash[:], *e.reactor.EnclaveInfo())
-	if err != nil {
-		msgVoteOracleRegistrationNo, err := makeOracleRegistrationVote(uniqueID, e.reactor.OracleAcc().GetAddress(), addressValue, types.VOTE_OPTION_NO, e.reactor.OraclePrivKey())
-		if err != nil {
-			return err
-		}
+	voteOption := verifyReportAndGetVoteOption(oracleRegistration, nodePubKeyHash, e)
 
-		txBytes, err := generateTxBytes(msgVoteOracleRegistrationNo, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), e.reactor.TxBuilder())
-		if err != nil {
-			return err
-		}
-
-		err = broadCastTx(e.reactor.GRPCClient(), txBytes)
-		if err != nil {
-			return err
-		}
-		return err
-	}
-
-	msgVoteOracleRegistrationYes, err := makeOracleRegistrationVote(uniqueID, e.reactor.OracleAcc().GetAddress(), addressValue, types.VOTE_OPTION_YES, e.reactor.OraclePrivKey())
-	if err != nil {
-		 return err
-	}
-
-	txBytes, err := generateTxBytes(msgVoteOracleRegistrationYes, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), e.reactor.TxBuilder())
+	msgVoteOracleRegistration, err := makeOracleRegistrationVote(uniqueID, e.reactor.OracleAcc().GetAddress(), addressValue, voteOption, e.reactor.OraclePrivKey())
 	if err != nil {
 		return err
 	}
 
-	err = broadCastTx(e.reactor.GRPCClient(), txBytes)
+	txBytes, err := generateTxBytes(msgVoteOracleRegistration, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), e.reactor.TxBuilder())
 	if err != nil {
+		return err
+	}
+
+	if err := broadCastTx(e.reactor.GRPCClient(), txBytes); err != nil {
 		return err
 	}
 
 	return nil
+}
+
+// verifyReportAndGetVoteOption validates the RemoteReport and returns the voting result according to the verification result.
+func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration, nodePubKeyHash [32]byte, e RegisterOracleEvent) types.VoteOption {
+	if err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, nodePubKeyHash[:], *e.reactor.EnclaveInfo()); err != nil {
+		log.Infof("failed to verification report. uniqueID(%s), address(%s), err(%v)", oracleRegistration.UniqueId, oracleRegistration.Address, err)
+		return types.VOTE_OPTION_NO
+	} else {
+		return types.VOTE_OPTION_YES
+	}
 }
 
 // makeOracleRegistrationVote makes a vote for oracle registration with VOTE_OPTION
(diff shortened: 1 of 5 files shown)


event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+}
+
+// makeOracleRegistrationVote makes a vote for oracle registration with VOTE_OPTION
+func makeOracleRegistrationVote(uniqueID, voterAddr, votingTargetAddr string, voteOption types.VoteOption, oraclePrivKey []byte) (*types.MsgVoteOracleRegistration, error) {
+	registrationVote := &types.OracleRegistrationVote{
+		UniqueId:               uniqueID,
+		VoterAddress:           voterAddr,
+		VotingTargetAddress:    votingTargetAddr,
+		VoteOption:             voteOption,
+		EncryptedOraclePrivKey: oraclePrivKey,
```
0xHansLee: The oracle key should be encrypted to the node public key of oracle registration.
Please encrypt it.

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	if err != nil {
+		return nil, err
+	}
+
+	sign, err := key.Sign(bytes)
+	if err != nil {
+		return nil, err
+	}
+
+	msgVoteOracleRegistrationNo := &types.MsgVoteOracleRegistration{
```
0xHansLee:
```suggestion
	msgVoteOracleRegistration := &types.MsgVoteOracleRegistration{
```

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+		OracleRegistrationVote: registrationVote,
+		Signature:              sign,
+	}
+
+	return msgVoteOracleRegistrationNo, nil
```
0xHansLee:
```suggestion
	return msgVoteOracleRegistration, nil
```

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+
+		txBytes, err := generateTxBytes(msgVoteOracleRegistrationNo, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), e.reactor.TxBuilder())
+		if err != nil {
+			return err
+		}
+
+		err = broadCastTx(e.reactor.GRPCClient(), txBytes)
+		if err != nil {
+			return err
+		}
```
0xHansLee: These codes are duplicated. I think we can share this code block with the case of `YES`

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	}
+
+	return txBytes, nil
+}
+
+// broadCastTx
+func broadCastTx(grpcClient *panacea.GrpcClient, txBytes []byte) error {
+	resp, err := grpcClient.BroadcastTx(txBytes)
+	if err != nil {
+		return err
```
0xHansLee, line 150:
```suggestion
		return fmt.Errorf("failed to broadcast MsgVoteOracleRegisration transaction: %w", err)
```
Let's returns more detailed error

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	}
+
+	if resp.TxResponse.Code != 0 {
+		return fmt.Errorf("register oracle vote failed: %v", resp.TxResponse.RawLog)
```
0xHansLee:
```suggestion
		return fmt.Errorf("register oracle vote transaction failed: %v", resp.TxResponse.RawLog)
```

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	}
```
0xHansLee, line 155: Like `register-oracle`, let's print the info log about transaction.
```suggestion
        }
        log.Infof("MsgVoteOracleRegistration transaction succeed. height(%v), hash(%s)", resp.TxResponse.Height, resp.TxResponse.TxHash)
```

0xHansLee (changes requested):
It is important to encrypt the oracle private key, so I request changes this PR.
Thank you!

event/register_oracle_event.go
```diff
@@ -51,9 +59,93 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+		txBytes, err := generateTxBytes(msgVoteOracleRegistrationNo, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), e.reactor.TxBuilder())
+		if err != nil {
+			return err
+		}
+
+		err = broadCastTx(e.reactor.GRPCClient(), txBytes)
+		if err != nil {
+			return err
+		}
 		return err
```
0xHansLee: When the remote report is invalid, if `generatingTxBytes` and `broadcastTx` are succeed, it should return `nil`, not `err`. Because it is a normal case of invalid oracle registration vote.
```suggestion
		return nil
```

service/service.go
```diff
@@ -46,13 +44,16 @@ func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, e
 		return nil, fmt.Errorf("failed to init subscriber: %w", err)
 	}
 
+	txBuilder := panacea.NewTxBuilder(grpcClient)
+
 	return &Service{
 		conf:          conf,
 		oracleAccount: oracleAccount,
-		oraclePrivKey: oraclePrivKey,
+		oraclePrivKey: oraclePrivKeyBz,
```
0xHansLee: Do you have any specific reason to use []byte of oracle private key, not `*btcec.PrivKey`?

gyuguen: In my opinion, it is better to use `*btcec.PrivKey` rather than `[]byte`.

gyuguen:
@inchori 
I made some improvements to this PR. #45 
If there is no problem, please merge. :)

event/register_oracle_event.go
gyuguen: Since this error occurred while executing `generateTxBytes()` or `broadCastTx()`, wouldn't it be appropriate to return an error?

event/register_oracle_event.go
gyuguen: Since this error occurred while executing `generateTxBytes()` or `broadCastTx()`, wouldn't it be appropriate to return an error?

0xHansLee: Right. However the error occurred from `generateTxBytes()` and `broadcastTx()` is checked if it is nil or not already.
If there is no error, I think, we should `return nil` even if the remote report is invalid.

gyuguen: Oh, I understand. In the current structure, we would vote `No` and then vote `Yes` again.

event/register_oracle_event.go
```diff
@@ -49,11 +58,90 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+
+// verifyReportAndGetVoteOption validates the RemoteReport and returns the voting result according to the verification result.
+func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration, nodePubKeyHash [32]byte, e RegisterOracleEvent) types.VoteOption {
+	if err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, nodePubKeyHash[:], *e.reactor.EnclaveInfo()); err != nil {
+		log.Infof("failed to verification report. uniqueID(%s), address(%s), err(%v)", oracleRegistration.UniqueId, oracleRegistration.Address, err)
+		return types.VOTE_OPTION_NO
+	} else {
+		return types.VOTE_OPTION_YES
+	}
+}
```
gyuguen: When I think about it again, changing it like this makes it clearer. :)
Because `nodePubKeyHash` is only used here.
```suggestion
func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration, enclaveInfo sgx.EnclaveInfo) types.VoteOption {
	nodePubKeyHash := sha256.Sum256(oracleRegistration.NodePubKey)
	if err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, nodePubKeyHash[:], enclaveInfo); err != nil {
		log.Infof("failed to verification report. uniqueID(%s), address(%s), err(%v)", oracleRegistration.UniqueId, oracleRegistration.Address, err)
		return types.VOTE_OPTION_NO
	} else {
		return types.VOTE_OPTION_YES
	}
}
```

inchori: Oh, yes. I'll change it.

service/service.go
inchori: It's because when make registration vote, the type of private key in `types.OracleRegistration`.
I'll check the method that convert type `*btcec.PrivKey` to `[]byte`.

0xHansLee: You can convert to `[]byte` using `btcec.Privkey.Serialize()`.

inchori: Thanks! I'll use `Serialize()` method.

event/register_oracle_event.go
inchori: Fixed in [a3ecc46](https://github.com/medibloc/panacea-doracle/pull/44/commits/a3ecc4648e60af34a710ae19e695c0fbecd2ca77).

event/register_oracle_event.go
```diff
@@ -46,11 +58,114 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	txBytes, err := txBuilder.GenerateSignedTxBytes(privKey, conf.Panacea.DefaultGasLimit, defaultFeeAmount, msgVoteOracleRegistration)
+	if err != nil {
+		return nil, err
+	}
+
+	return txBytes, nil
+}
+
+// broadCastTx
+func broadCastTx(grpcClient *panacea.GrpcClient, txBytes []byte) error {
```
youngjoon-lee:
```suggestion
// broadcastTx
func broadcastTx(grpcClient *panacea.GrpcClient, txBytes []byte) error {
```

event/register_oracle_event.go
```diff
@@ -46,11 +58,114 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+	key := secp256k1.PrivKey{
+		Key: oraclePrivKey,
+	}
+
+	bytes, err := registrationVote.Marshal()
+	if err != nil {
+		return nil, err
+	}
+
+	sign, err := key.Sign(bytes)
```
youngjoon-lee:
```suggestion
	sig, err := key.Sign(bytes)
```

event/register_oracle_event.go
```diff
@@ -46,11 +58,114 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
+
+	return nil
+}
+
+// verifyReportAndGetVoteOption validates the RemoteReport and returns the voting result according to the verification result.
+func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration, e RegisterOracleEvent) types.VoteOption {
+	nodePubKeyHash := sha256.Sum256(oracleRegistration.NodePubKey)
+
+	if err := sgx.VerifyRemoteReport(oracleRegistration.NodePubKeyRemoteReport, nodePubKeyHash[:], *e.reactor.EnclaveInfo()); err != nil {
+		log.Infof("failed to verification report. uniqueID(%s), address(%s), err(%v)", oracleRegistration.UniqueId, oracleRegistration.Address, err)
```
youngjoon-lee:
```suggestion
		log.Warnf("failed to verification report. uniqueID(%s), address(%s), err(%v)", oracleRegistration.UniqueId, oracleRegistration.Address, err)
```

Or `Errorf`.

It would be good to not use INFO in this case, so that we can recognize issues quickly when monitoring logs.

I suggested the `Warnf` because the error can occur not only when something went wrong in our process, but also when the submitted remote report is invalid. But, IMHO, both are good. Please choose what you want.

inchori: I understand. For now, I'll just change it to `Warnf`.

gyuguen (approved):
LGTM 👍

0xHansLee (approved):
LGTM.
But as I said offline, we need to validate the trusted block info.

inchori:
This branch was approved, so I'll merge this into the base branch. And the validating the trusted block info will be implemented the another branch.

PR #48, opened 2022-09-08
feat: add query client to service

0xHansLee:
Added query client to service and refactored the `RegisterOracleEvent`'s event handler accordingly.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -44,12 +44,20 @@ func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, e
 
 	grpcClient, err := panacea.NewGrpcClient(conf)
 	if err != nil {
+		if err := queryClient.Close(); err != nil {
+			log.Warn(err)
+		}
 		return nil, fmt.Errorf("failed to create a new gRPC client: %w", err)
 	}
 
 	subscriber, err := event.NewSubscriber(conf.Panacea.RPCAddr)
 	if err != nil {
-		_ = grpcClient.Close()
+		if err := queryClient.Close(); err != nil {
+			log.Warn(err)
+		}
+		if err := grpcClient.Close(); err != nil {
+			log.Warn(err)
+		}
 		return nil, fmt.Errorf("failed to init subscriber: %w", err)
 	}
 
--- a/service/service.go
+++ b/service/service.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"context"
 	"fmt"
 	"github.com/btcsuite/btcd/btcec"
 	"github.com/medibloc/panacea-doracle/config"
@@ -18,8 +19,9 @@ type Service struct {
 	oracleAccount *panacea.OracleAccount
 	oraclePrivKey *btcec.PrivateKey
 
-	grpcClient *panacea.GrpcClient
-	subscriber *event.PanaceaSubscriber
+	queryClient *panacea.QueryClient
+	grpcClient  *panacea.GrpcClient
+	subscriber  *event.PanaceaSubscriber
 }
 
 func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, error) {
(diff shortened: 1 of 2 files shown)


inchori (approved):
LGTM 👍

service/service.go
```diff
@@ -35,14 +37,19 @@ func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, e
...
+
 	grpcClient, err := panacea.NewGrpcClient(conf)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create a new gRPC client: %w", err)
 	}
 
 	subscriber, err := event.NewSubscriber(conf.Panacea.RPCAddr)
 	if err != nil {
-		// TODO: close grpcClient
+		_ = grpcClient.Close()
```
gyuguen: If an error occurs while closing grpcClient, how about outputting a warning error here?

service/service.go
```diff
@@ -62,7 +69,9 @@ func (s *Service) StartSubscriptions(events ...event.Event) error {
 }
 
 func (s *Service) Close() error {
-	// TODO close query client
+	if err := s.queryClient.Close(); err != nil {
```
gyuguen, line 72: Oh, it's already outputting a warning log. If so, it looks like you don't need a return value.

gyuguen (approved):
LGTM 👍

service/service.go
0xHansLee: Should we `panic(err)` when error is not nil from closing gRPC client?

gyuguen: I think it would be better to print an error log rather than panic so that you can see what kind of error occurred.

```suggestion
		if err := grpcClient.Close(); err != nil {
		    log.Warn(err)
		}
```

0xHansLee: I think we may have to close query client as well.

0xHansLee: Ok. I see. I'll make changes as you suggested!

gyuguen: > I think we may have to close query client as well.

Agree.

audtlr24 (approved):
Good! 👍

service/service.go
0xHansLee: Wait, do we have to close here?
Actually the caller deferred close the service, so all the resources will close.

audtlr24: I think `queryClient` and `grpcClient` is closed at `svc.Close()`. Is it right?

0xHansLee: Yes that's what I thought. So I think we don't have to close each resource when error occurred.
The caller should close the service when the caller creates a new service.
Am I right @gyuguen?

gyuguen: Yes. However, since the place where the service is created is implemented like this, it seems that the order needs to be changed.

```go
svc, err := service.New(conf, oracleAccount)
if err != nil {
	return fmt.Errorf("failed to create service: %w", err)
}
defer svc.Close()
```

youngjoon-lee: If any subcomponent isn't initialized successfully in the `service.New()`, subcomponents that were already initialized in the `service.New()` must be closed explicitly before returning the error from `service.New()`, as Gyuguen said.

Also, an error from any `Close()` function shouldn't stop any remaining tasks. It is good enough just to print an WARN log. After that, we should keep going for remaining tasks, such as closing other subcomponents or returning an error to the caller.
Errors from `Close()` are usually from closing TCP sockets which were already closed for some reason, from disk I/O errors, or from deadlocks in the subcomponents. Those errors are generally unrecoverable. It means that there is nothing for us to do with those errors. In fact, most of errors from `Close()` are not critical in general.

0xHansLee: I got it, thank you.
When error occurs in creating new service, the resources are explicitly closed as you said.

service/service.go
```diff
@@ -44,12 +44,20 @@ func New(conf *config.Config, oracleAccount *panacea.OracleAccount) (*Service, e
 
 	grpcClient, err := panacea.NewGrpcClient(conf)
 	if err != nil {
+		if err := queryClient.Close(); err != nil {
```
audtlr24, line 47: I think there is typo. It seemed to be `grpcClient`.

0xHansLee: It's not typo.
If an error occurs in creating new gRPC client, we need to close the resources previously initialized.

audtlr24: Ah, okay. I understand now. Thanks! 👍

PR #50, opened 2022-09-13
feat: add validate trusted block info when handling register oracle event

inchori:
- Validating trusted block info when handling register oracle event.
- When the block hash different between blockchain and node that wants to register oracle, it returns an error.


gyuguen:
@inchori 
I think it's better to compare Hash with TrustedBlock obtained from LightClient.
Please review this PR. #53

gyuguen (approved):
LGTM 👍

audtlr24 (approved):
LGTM!

0xHansLee (approved):

PR #51, opened 2022-09-13
feat: add trustedBlock info at gen-oracle-key when creating the genesis oracle

audtlr24:
To resolve the #49 issue, fix command `gen-oracle-key`.

- get trusted block info and initiate `QueryClient`

### DIFF
--- a/cmd/doracled/cmd/gen_oracle_key.go
+++ b/cmd/doracled/cmd/gen_oracle_key.go
@@ -82,7 +82,7 @@ So please be cautious in using this command.`,
 			if err != nil {
 				return err
 			}
-			if isGenesisOracle == true {
+			if isGenesisOracle {
 				// get trusted block information
 				trustedBlockInfo, err := getTrustedBlockInfo(cmd)
 				if err != nil {
--- a/cmd/doracled/cmd/gen_oracle_key.go
+++ b/cmd/doracled/cmd/gen_oracle_key.go
@@ -77,32 +77,26 @@ So please be cautious in using this command.`,
 				return err
 			}
 
-			// If FlagIsGenesisOracle is true, create a QueryClient
-			isGenesisOracle, err := cmd.Flags().GetBool(flags.FlagIsGenesisOracle)
+			// get trusted block information
+			trustedBlockInfo, err := getTrustedBlockInfo(cmd)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to get trusted block info: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


cmd/doracled/cmd/gen_oracle_key.go
```diff
@@ -74,9 +77,33 @@ So please be cautious in using this command.`,
...
+				if err != nil {
+					return fmt.Errorf("failed to initialize QueryClient: %w", err)
+				}
+				defer queryClient.Close()
+			}
+
 			return nil
 		},
 	}
+	cmd.Flags().Bool(flags.FlagIsGenesisOracle, false, "Check if it is the genesis oracle creation")
```
gyuguen: I think this value is not needed.

This is because the `gen-oracle-key` command is executed only once by the first oracle.

In other words, `flags.FlagTrustedBlockHeight` and `flags.FlagTrustedBlockHash` would be required values.

audtlr24: I'm a little confused. Aren't all oracles running `gen-oracle-key` to generate `nodePrivKey`?

gyuguen: No. Generate `OraclePrivKey`. :)

audtlr24: Ah, sorry. I confused. Okay, I will fix it 👍

0xHansLee (approved):
I'm not sure the trusted block info is required in this cmd, but let's keep it this way for now.

gyuguen:
> I'm not sure the trusted block info is required in this cmd, but let's keep it this way for now.

Yes. I'm not sure either.
But I can't think of which other cmd to add right now. :)

gyuguen (approved):
LGTM 👍

0xHansLee:
> >I'm not sure the trusted block info is required in this cmd, but let's keep it this way for now.
> 
> Yes. I'm not sure either.
> But I can't think of which other cmd to add right now. :)

I meant it could be optional flags. I also think it's good to add to this cmd.

gyuguen:
Aha. The part you were worried about is whether to receive it as a flag or the content of the argument?

0xHansLee:
No I think it's good enough to accept trusted block info using flags.

I was just not sure if that value should be required, (not optional). But it maybe due to the name of cmd, `gen-oracle-key`, as youngjoon said in the issue(#49). It's not a big deal :)

PR #54, opened 2022-09-14
fix: change encrypt/decrpyt method to AES256 cryptography

### LINKED ISSUE #52: bug: incorrect "OraclePrivKey" encrypt
We use the encryption and decryption method provided by btcec.

**crypto/encrypt.go**
```go


// Encrypt encrypts data using a secp256k1 public key (ECIES)
func Encrypt(pubKey *btcec.PublicKey, data []byte) ([]byte, error) {
	return btcec.Encrypt(pubKey, data)
}

// Decrypt decrypts data using a secp256k1 private key (ECIES)
func Decrypt(privKey *btcec.PrivateKey, data []byte) ([]byte, error) {
	return btcec.Decrypt(privKey, data)
}
```

The encryption method used here internally generates a new key every time. Therefore, even if the input is the same, the output is always different.

So you should refer to this code.
https://github.com/medibloc/doracle-poc/blob/with-panacea/pkg/secp256k1/encrypt.go#L14

example)
https://github.com/medibloc/doracle-poc/blob/with-panacea/pkg/secp256k1/encrypt_test.go#L37

inchori:
- To resolve [#52](https://github.com/medibloc/panacea-doracle/issues/52), change encrypt/decrypt method to AES256 when vote or join to oracle.
- I mark as TODO for add nonce in encrypt oraclePrivKey with shareKey when make `oracleRegistrationVote`.

### DIFF
--- a/cmd/doracled/cmd/get_oracle_key.go
+++ b/cmd/doracled/cmd/get_oracle_key.go
@@ -3,8 +3,10 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"errors"
 	"fmt"
+	"io"
 
 	"github.com/btcsuite/btcd/btcec"
 	oracletypes "github.com/medibloc/panacea-core/v2/x/oracle/types"
@@ -79,7 +81,14 @@ func getOracleKeyCmd() *cobra.Command {
 				return errors.New("the existing node key is different from the one used in oracle registration. if you want to re-request RegisterOracle, delete the existing node_priv_key.sealed file and rerun register-oracle cmd")
 			}
 
-			return getOraclePrivKey(conf, oracleRegistration, nodePrivKey, queryClient)
+			// TODO: This will be changed to the nonce that is field of oracleRegistration
+			nonce := make([]byte, 12)
+			size, err := io.ReadFull(rand.Reader, nonce)
+			if size == 0 {
+				return err
+			}
+
+			return getOraclePrivKey(conf, oracleRegistration, nodePrivKey, nonce, queryClient)
 		},
 	}
 	cmd.Flags().Uint32P(flags.FlagAccNum, "a", 0, "Account number of oracle")
@@ -89,7 +98,7 @@ func getOracleKeyCmd() *cobra.Command {
 }
 
 // getOraclePrivKey handles OracleRegistration differently depending on the status of oracle registration
-func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.OracleRegistration, nodePrivKey *btcec.PrivateKey, queryClient *panacea.QueryClient) error {
+func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.OracleRegistration, nodePrivKey *btcec.PrivateKey, nonce []byte, queryClient *panacea.QueryClient) error {
 	switch oracleRegistration.Status {
 	case oracletypes.ORACLE_REGISTRATION_STATUS_VOTING_PERIOD:
 		return errors.New("voting is currently in progress")
@@ -117,7 +126,7 @@ func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.Oracl
 
 		shareKey := crypto.ShareKey(nodePrivKey, oraclePubKey)
 
-		oraclePrivKey, err := crypto.DecryptWithAES256(shareKey, oracleRegistration.EncryptedOraclePrivKey)
+		oraclePrivKey, err := crypto.DecryptWithAES256(shareKey, nonce, oracleRegistration.EncryptedOraclePrivKey)
 		if err != nil {
 			return fmt.Errorf("failed to decrypt the encrypted oracle private key: %w", err)
 		}
--- a/cmd/doracled/cmd/get_oracle_key.go
+++ b/cmd/doracled/cmd/get_oracle_key.go
@@ -36,7 +36,7 @@ func getOracleKeyCmd() *cobra.Command {
 
 			ctx := context.Background()
 
-			// if there is a node key already, return error
+			// if the node key does not exist, return error
 			nodePrivKeyPath := conf.AbsNodePrivKeyPath()
 			if !tos.FileExists(nodePrivKeyPath) {
 				return errors.New("no node_priv_key.sealed file")
--- a/cmd/doracled/cmd/get_oracle_key.go
+++ b/cmd/doracled/cmd/get_oracle_key.go
@@ -106,7 +106,7 @@ func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.Oracl
 			return errors.New("the oracle private key already exists")
 		}
 
-		shareKey := crypto.ShareKey(nodePrivKey, oraclePubKey)
+		shareKey := crypto.SharedKey(nodePrivKey, oraclePubKey)
 
 		oraclePrivKey, err := crypto.DecryptWithAES256(shareKey, oracleRegistration.Nonce, oracleRegistration.EncryptedOraclePrivKey)
 		if err != nil {
--- a/cmd/doracled/cmd/get_oracle_key.go
+++ b/cmd/doracled/cmd/get_oracle_key.go
@@ -79,7 +79,12 @@ func getOracleKeyCmd() *cobra.Command {
 				return errors.New("the existing node key is different from the one used in oracle registration. if you want to re-request RegisterOracle, delete the existing node_priv_key.sealed file and rerun register-oracle cmd")
 			}
 
-			return getOraclePrivKey(conf, oracleRegistration, nodePrivKey, oracleRegistration.Nonce, queryClient)
+			oraclePublicKey, err := queryClient.GetOracleParamsPublicKey()
+			if err != nil {
+				return err
+			}
+
+			return getOraclePrivKey(conf, oracleRegistration, nodePrivKey, oraclePublicKey)
 		},
 	}
 	cmd.Flags().Uint32P(flags.FlagAccNum, "a", 0, "Account number of oracle")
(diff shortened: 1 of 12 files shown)


event/register_oracle_event.go
```diff
@@ -92,12 +93,11 @@ func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration,
 
 // makeOracleRegistrationVote makes a vote for oracle registration with VOTE_OPTION
 func makeOracleRegistrationVote(uniqueID, voterAddr, votingTargetAddr string, voteOption types.VoteOption, oraclePrivKey []byte, nodePubKey []byte) (*types.MsgVoteOracleRegistration, error) {
-	pubKey, err := btcec.ParsePubKey(nodePubKey, btcec.S256())
-	if err != nil {
-		return nil, err
-	}
+	privKey, pubKey := crypto.PrivKeyFromBytes(oraclePrivKey)
+	shareKey := crypto.ShareKey(privKey, pubKey)
```
0xHansLee: `nodePubKey` should be used for shareKey, not oracle public key :)

cmd/doracled/cmd/get_oracle_key.go
```diff
@@ -100,8 +101,20 @@ func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.Oracl
...
+		oraclePubKeyPath := conf.AbsOraclePubKeyPath()
+		unsealOraclePubKey, err := sgx.UnsealFromFile(oraclePubKeyPath)
+		if err != nil {
+			return fmt.Errorf("failed to unseal from file: %w", err)
+		}
+
+		oraclePubKey, err := btcec.ParsePubKey(unsealOraclePubKey, btcec.S256())
+		if err != nil {
+			return fmt.Errorf("failed to parse oraclePubKey: %w", err)
+		}
```
0xHansLee: As you can probably see the `conf.AbsOraclePubKeyPath()`, the oracle pubkey file(default name of `oracle_pub_key.json`) is used to store oracle public key and its remote report in `gen-oracle-key` cmd.

In general, the oracle to register probably don't have the file, thus how about to get the oracle public key from Panacea using query client?

gyuguen:
Please add test code.
As I expected, encryption/decryption in the test code will fail.

cmd/doracled/cmd/get_oracle_key.go
inchori: Ah yes, I understand. The `oracle_pub_key.json` will be generated when using `gen-oracle-key` cmd. So I think you're right. I'll change it.

inchori: Fixed in [bd2279e](https://github.com/medibloc/panacea-doracle/pull/54/commits/bd2279e57d2a4307de9ce396325a2a35d9d43e16).

cmd/doracled/cmd/get_oracle_key.go
```diff
@@ -78,7 +81,14 @@ func getOracleKeyCmd() *cobra.Command {
 				return errors.New("the existing node key is different from the one used in oracle registration. if you want to re-request RegisterOracle, delete the existing node_priv_key.sealed file and rerun register-oracle cmd")
 			}
 
-			return getOraclePrivKey(conf, oracleRegistration, nodePrivKey)
+			// TODO: This will be changed to the nonce that is field of oracleRegistration
```
gyuguen: This change is reflected in panacea-core. (https://github.com/medibloc/panacea-core/pull/404)
Modify it using the master branch of panacea-core.

inchori: BTW, I think we need to add `Nonce` field in `NewMsgRegisterOracle` in [x/oracle/type/message_oracle.go](https://github.com/medibloc/panacea-core/blob/2c0f80b0f08e2ba3db98593ddeb384931618dae2/x/oracle/types/message_oracle.go#L10).

cmd/doracled/cmd/get_oracle_key.go
```diff
@@ -88,7 +98,7 @@ func getOracleKeyCmd() *cobra.Command {
 }
 
 // getOraclePrivKey handles OracleRegistration differently depending on the status of oracle registration
-func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.OracleRegistration, nodePrivKey *btcec.PrivateKey) error {
+func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.OracleRegistration, nodePrivKey *btcec.PrivateKey, nonce []byte, queryClient *panacea.QueryClient) error {
```
0xHansLee: The nonce would be included in `oracleRegistration` as you can see [here](https://github.com/medibloc/panacea-core/blob/2c0f80b0f08e2ba3db98593ddeb384931618dae2/proto/panacea/oracle/v2alpha2/oracle.proto#L71)

cmd/doracled/cmd/get_oracle_key.go
gyuguen: Yes, I created this PR.(https://github.com/medibloc/panacea-core/pull/405)

crypto/encrypt.go
```diff
@@ -11,3 +17,50 @@ func Encrypt(pubKey *btcec.PublicKey, data []byte) ([]byte, error) {
 func Decrypt(privKey *btcec.PrivateKey, data []byte) ([]byte, error) {
 	return btcec.Decrypt(privKey, data)
 }
+
+func EncryptWithAES256(key, nonce, data []byte) ([]byte, error) {
```
0xHansLee:
```suggestion
func EncryptWithAES256(secretKey, nonce, data []byte) ([]byte, error) {
```
I suggested to rename the `key` to `secretKey`.

cmd/doracled/cmd/get_oracle_key.go
```diff
@@ -100,10 +110,25 @@ func getOraclePrivKey(conf *config.Config, oracleRegistration *oracletypes.Oracl
 			return errors.New("the oracle private key already exists")
 		}
 
-		// else, get encryptedOraclePrivKey from Panacea and decrypt and SealToFile it
-		oraclePrivKey, err := crypto.Decrypt(nodePrivKey, oracleRegistration.EncryptedOraclePrivKey)
+		oraclePublicKey, err := queryClient.GetOracleParamsPublicKey()
```
0xHansLee: I think it would be better to get the `oraclePubKey` outside this function.

event/register_oracle_event.go
```diff
@@ -102,12 +103,16 @@ func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration,
...
+
 	pubKey, err := btcec.ParsePubKey(nodePubKey, btcec.S256())
 	if err != nil {
 		return nil, err
 	}
 
-	encryptedOraclePrivKey, err := crypto.Encrypt(pubKey, oraclePrivKey)
+	shareKey := crypto.ShareKey(privKey, pubKey)
+	// TODO: Nonce will be added.
+	encryptedOraclePrivKey, err := crypto.EncryptWithAES256(shareKey, nil, oraclePrivKey)
```
0xHansLee: The nonce may be in the `oracleRegistration` obtained using query client in line 57. I think you can add it now :)

panacea/query_client.go
```diff
@@ -316,3 +318,12 @@ func (q QueryClient) GetOracleRegistration(oracleAddr, uniqueID string) (*oracle
 func (q QueryClient) GetLightBlock(height int64) (*tmtypes.LightBlock, error) {
 	return q.lightClient.TrustedLightBlock(height)
 }
+
+func (q QueryClient) GetOracleParamsPublicKey() ([]byte, error) {
```
0xHansLee: Why don't we return the parsed oracle public key instead of byte array?
But it's not important for now. I just wonder your thoughts.

crypto/encrypt.go
inchori: Ah yes, I agree

panacea/query_client.go
inchori: It doesn't have a big reason. But I think that put the parsing the byte of oracle public key to this function would be more better. So I can remove error handling for parsing public key in `get_oracle_key.go`.

inchori: Fixed in [a0fe48a](https://github.com/medibloc/panacea-doracle/pull/54/commits/a0fe48a929c83780c86a7930b28e5a5e62dc6885).

panacea/query_client.go
```diff
@@ -319,11 +320,19 @@ func (q QueryClient) GetLightBlock(height int64) (*tmtypes.LightBlock, error) {
 	return q.lightClient.TrustedLightBlock(height)
 }
 
-func (q QueryClient) GetOracleParamsPublicKey() ([]byte, error) {
+func (q QueryClient) GetOracleParamsPublicKey() (*btcec.PublicKey, error) {
 	oraclePubKeyBz, err := q.GetStoreData(context.Background(), paramstypes.StoreKey, append(append([]byte(oracletypes.StoreKey), '/'), oracletypes.KeyOraclePublicKey...))
 	if err != nil {
 		return nil, err
 	}
+	if oraclePubKeyBz == nil {
```
inchori, line 328: For now, I just put an error handling for nil value of oracle public key when querying oracle params in panacea-core.

panacea/query_client.go
```diff
@@ -316,3 +319,20 @@ func (q QueryClient) GetOracleRegistration(oracleAddr, uniqueID string) (*oracle
...
 }
+
+func (q QueryClient) GetOracleParamsPublicKey() (*btcec.PublicKey, error) {
+	oraclePubKeyBz, err := q.GetStoreData(context.Background(), paramstypes.StoreKey, append(append([]byte(oracletypes.StoreKey), '/'), oracletypes.KeyOraclePublicKey...))
+	if err != nil {
+		return nil, err
+	}
+	if oraclePubKeyBz == nil {
+		return nil, errors.New("the oracle public key's value is nil")
+	}
```
inchori, line 330: For now, I just put an error handling for nil value of oracle public key when querying oracle params in panacea-core.

gyuguen (approved):
LGTM 👍

0xHansLee (approved):

crypto/key.go
```diff
@@ -27,3 +28,7 @@ func GeneratePrivateKeyFromMnemonic(mnemonic string, coinType, accNum, index uin
 
 	return hd.DerivePrivateKeyForPath(master, ch, hdPath)
 }
+
+func ShareKey(priv *btcec.PrivateKey, pub *btcec.PublicKey) []byte {
```
youngjoon-lee:
```suggestion
func SharedKey(priv *btcec.PrivateKey, pub *btcec.PublicKey) []byte {
```

youngjoon-lee (approved):

PR #55, opened 2022-09-14
refactor: remove unused code

0xHansLee:
All the gRPC queries are replaced to light client query, so removed unused code.


gyuguen (approved):
LGTM 👍 

We see that we can use grpcClient for non-critical values.
However, we haven't decided whether to use grpcCleint yet, so it looks like we can delete it.

audtlr24 (approved):
lgtm!

PR #56, opened 2022-09-14
ci: add integration test suite

youngjoon-lee:
Thanks to [ory/dockertest](https://github.com/ory/dockertest), I added a `integration.TestSuite` to run a `panacea-core` docker container before starting some unit tests. The docker container will be terminated when each unit test is finalized. This way is better than what we've done in the [CI](https://github.com/medibloc/panacea-js/blob/601fe765617fa3a1de294c0d51e820c2309a04ed/.github/workflows/ci.yml#L35) of `panacea-js`.

Because the [Dockerfile](https://github.com/medibloc/panacea-core/blob/3f5922acb220bb4856ae02f3111acf9a62d92423/Dockerfile#L30) in the `panacea-core` repo was written as general as possible, I've made each test inject an [init_script](https://github.com/medibloc/panacea-doracle/compare/ci/na/docker?expand=1#diff-5d8f8235fc9a7f2dd769576d32d6a1570a4aa8dfaa2c9af1abfa03160affe007R39) to the docker container in order to initialize/start a chain. This would be also flexible to configure chains in different ways as each unit test requires.

I used this TestSuite for the `query_client_test.go` with some refactoring. Previously, that has contained too many tests. For example, that doesn't need to be run in SGX. That can be tested outside SGX with a MemDB because our purpose is testing if `QueryClient` work well with light client proof. The `SGXLevelDB` must be tested in the `github.com/medibloc/panacea-doracle/store/sgxleveldb` package.
(Actually, I haven't figure out how to run a `panacea-core` docker container when running tests in SGX. So, let's split tests clearly so that we don't mix everything in one test.)


integration/rest.go
```diff
@@ -0,0 +1,53 @@
...
+		} `json:"header"`
+	} `json:"block"`
+}
+
+func QueryLatestBlock(endpoint string) ([]byte, int64, error) {
+	resp, err := http.Get(fmt.Sprintf("%s/blocks/latest", endpoint))
+	if err != nil {
+		return nil, 0, fmt.Errorf("failed to execute HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
```
youngjoon-lee, line 27: I'm thinking of changing this logic to use gRPC instead of REST, but let's go with this code for now. I'm gonna find a nicer way when designing the panacea-e2e.

integration/suite.go
```diff
@@ -0,0 +1,114 @@
...
+	suite.dktPool = dktPool
+}
+
+func (suite *TestSuite) SetupTest() {
+	var err error
+
+	suite.dktResource, err = suite.dktPool.RunWithOptions(
+		&dockertest.RunOptions{
+			Repository: "ghcr.io/medibloc/panacea-core",
+			Tag:        "master",
```
youngjoon-lee, line 54: This docker image can be found in the https://github.com/medibloc/panacea-core/pkgs/container/panacea-core (published by https://github.com/medibloc/panacea-core/blob/master/.github/workflows/docker-publish.yml)

panacea/query_client_test.go
```diff
@@ -2,285 +2,121 @@ package panacea
...
-	}()
-	wg.Wait()
-
+	return trustedBlockInfo, conf
 }
-
-// The test below is commented out because it is a test targeting local panacea.
-
-// Test for GetOracleRegistration function.
-//func TestGetOracleRegistration(t *testing.T) {
```
youngjoon-lee, line 209: We don't need to test all helper functions defined in the `query_client.go`. It's good enough to just test the core logic which may be `GetStoreData()`. But, instead, I left the `TestGetAccount()` as it was.

.github/workflows/ci.yaml
```diff
@@ -34,7 +34,21 @@ jobs:
...
+        id: go
+
+      - name: Checkout code into the Go module directory
+        uses: actions/checkout@v2
+
       - name: Lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.30
+          version: v1.30
```
youngjoon-lee, line 54: Since the `golangci-lint` eats lot of memory, I'm gonna run this job on Ubuntu hosted by GitHub, not self-hosted.

.github/workflows/ci.yaml
```diff
@@ -34,7 +34,22 @@ jobs:
...
+
+      - name: Checkout code into the Go module directory
+        uses: actions/checkout@v3
+
       - name: Lint
         uses: golangci/golangci-lint-action@v3
         with:
-          version: v1.30
+          version: latest
+          args: --timeout=5m
```
youngjoon-lee, line 55: Since the `golangci-lint` eats lot of memory, I'm gonna run this job on Ubuntu hosted by GitHub, not self-hosted.

gyuguen (approved):
LGTM Thanks!

audtlr24 (approved):
Thanks, LGTM!

0xHansLee (approved):
lgtm 👍

inchori (approved):
lgtm

PR #60, opened 2022-09-15
fix: add nonce when register oracle

### LINKED ISSUE #57: feat: added to include nonce when executing 'register-oracle' command
We defined the nonce value used in `AES-GCM` to be set when oracle make a registration request.
Therefore, when executing the `register-oracle` command, the `nonce` value must also be created and modifeid to be inserted.

inchori:
- Add nonce field in `NewMsgRegisterOracle` when register oracle cmd.

### DIFF
--- a/cmd/doracled/cmd/register_oracle.go
+++ b/cmd/doracled/cmd/register_oracle.go
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


0xHansLee (approved):

gyuguen (approved):
LGTM 👍

PR #66, opened 2022-09-16
fix: `SgxLevelDB.Get()` should return a `nil` value if the key is not found

### LINKED ISSUE #65: `SgxLevelDB.Get()` should return a `nil` value if the key is not found
Because `SgxLevelDB` implements the interface [tmdb.DB](https://github.com/tendermint/tm-db/blob/b599fa5d668d5a68812ec5ebeccb9eb05235bb3a/types.go#L21), it should conform to the spec of `tmdb.DB` completely.

The `tmdb.DB.Get()` should return a `nil` value if the key is not found. 
https://github.com/tendermint/tm-db/blob/b599fa5d668d5a68812ec5ebeccb9eb05235bb3a/types.go#L22
The `tmdb.GoLevelDB.Get()` (which implements `tmdb.DB`) was also implement like that.
https://github.com/tendermint/tm-db/blob/b599fa5d668d5a68812ec5ebeccb9eb05235bb3a/goleveldb.go#L50

This issue is slightly related to the issue https://github.com/medibloc/panacea-doracle/issues/62.

youngjoon-lee:
Closes #65

### DIFF
--- a/store/sgxleveldb/db.go
+++ b/store/sgxleveldb/db.go
@@ -42,11 +42,14 @@ func (sdb *SgxLevelDB) Get(key []byte) ([]byte, error) {
 	val, err := sdb.GoLevelDB.Get(key)
 	if err != nil {
 		return nil, err
+	} else if val == nil {
+		return nil, nil
 	}
+
 	log.Debug("unsealing after reading from leveldb")
 	unsealedVal, err := sgx.Unseal(val, true)
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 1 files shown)


gyuguen (approved):
LGTM 👍

audtlr24 (approved):
lgtm!

0xHansLee (approved):
👍

inchori (approved):
lgtm

PR #68, opened 2022-09-16
fix: handle the cases when `rpcClient.ABCIQueryWithOptions()` returns invalid responses

### LINKED ISSUE #67: Handle the case when `ABCIQueryWithOptions()` returns NotFound
Currently, we try to merkle proof unless `q.rpcClient.ABCIQueryWithOptions() returns an error.
https://github.com/medibloc/panacea-doracle/blob/aa3ef1426d821ea80b78a18e9c37b368ab2be972/panacea/query_client.go#L224

It looks reasonable, but @audtlr24 found that the `result.Response.Value` or `result.Response.ProofOps` can be `nil` if the key is not found, even though this behavior wasn't written well in the [documentation](https://github.com/tendermint/tendermint/blob/0c96f0b434c1d9f23883d1f2291135830a7a5290/rpc/client/interface.go#L55) of the interface `ABCIClient`.

Thus, we should handle the NotFound case properly.

youngjoon-lee:
Closes #67 

Initially, I tried to fix this code as below:
```go
result, err := q.rpcClient.ABCIQueryWithOptions(ctx, fmt.Sprintf("/store/%s/key", storeKey), key, option)
if err != nil {
    return nil, err
} else if result.Response.Value == nil || result.Response.ProofOps == nil {
    return nil, fmt.Errorf("ABCIQuery returned NotFound")
}
```
But, I've found there are still many edge cases that we should handle. I referred to the code in Tendermint: https://github.com/tendermint/tendermint/blob/0c96f0b434c1d9f23883d1f2291135830a7a5290/light/rpc/client.go#L139.
I was thinking of just using Tendermint's implementation directly, but couldn't figure it if it would work or not in our codebase. So, for now, I just copied their code blocks to our codebase.

### DIFF
--- a/panacea/query_client.go
+++ b/panacea/query_client.go
@@ -286,6 +286,9 @@ func (q QueryClient) abciQueryWithOptions(ctx context.Context, path string, data
 	if len(resp.Key) == 0 {
 		return nil, errors.New("empty key")
 	}
+	if len(resp.Value) == 0 {
+		return nil, errors.New("empty value")
+	}
 	if opts.Prove && (resp.ProofOps == nil || len(resp.ProofOps.Ops) == 0) {
 		return nil, errors.New("no proof ops")
 	}
--- a/panacea/query_client.go
+++ b/panacea/query_client.go
@@ -290,7 +290,7 @@ func (q QueryClient) abciQueryWithOptions(ctx context.Context, path string, data
 		return nil, errors.New("no proof ops")
 	}
 	if resp.Height <= 0 {
-		return nil, fmt.Errorf("negative or zero height")
+		return nil, errors.New("negative or zero height")
 	}
 
 	return res, nil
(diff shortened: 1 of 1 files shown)


gyuguen (approved):
Good. I see no problem. :)

audtlr24:
I think it would be better to handle error when value is nil. If there is no error handling, it seems that an error will occur in `VerifyMembership` during the existing merkle-proof. I think we can apply the `VerifyValue` part at the bottom of the referenced code, and change merkle-proof part to it. What do you think?

youngjoon-lee:
@audtlr24 
> I think it would be better to handle error when value is nil. If there is no error handling, it seems that an error will occur in VerifyMembership during the existing merkle-proof. 

Makes sense. I added an empty value handling: [f474071](https://github.com/medibloc/panacea-doracle/pull/68/commits/f4740718775c95f042f51213a3242bc769741955). Instead of `if resp.Value == nil`, I used `if len(resp.Value) == 0` because of https://github.com/golang/go/blob/d8220ac32fc73e69aa6d58890b6b5277d277825f/src/builtin/builtin.go#L165.

> I think we can apply the VerifyValue part at the bottom of the referenced code, and change merkle-proof part to it. What do you think?

I'm not sure if it works in our codebase. I think it's better to do it in a separated PR and test it in detail. Let's keep this PR small.

audtlr24 (approved):
Okay, I will test it and if the test results are good, I will open another PR.

PR #70, opened 2022-09-16
fix: handle base64 values returned by Panacea properly

### LINKED ISSUE #69: Handle base64 values returned by Panacea properly
When we query the params of `x/oracle` from Panacea, some of values returned are base64-encoded. So, we should decode them before using them.
https://github.com/medibloc/panacea-doracle/blob/aa3ef1426d821ea80b78a18e9c37b368ab2be972/panacea/query_client.go#L324

youngjoon-lee:
Close #69 

But, this code won't work well before resolving https://github.com/medibloc/panacea-core/issues/406, because the returned `[]byte` from panacea-core is actually `[]byte("\"value\"")` (with double quotations) for some reason. I guess it's the way how Protobuf encodes a `string` value to a `bytes`.
If you run this code without changing anything in panacea-core, you will get an error:
```
illegal base64 data at input byte 0
```
, because the first character `"` is not in the base64 charset.

To resolve this issue, I've made the PR https://github.com/medibloc/panacea-core/pull/407.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
 	github.com/cosmos/go-bip39 v1.0.0
 	github.com/cosmos/ibc-go/v2 v2.0.3
 	github.com/edgelesssys/ego v1.0.0
-	github.com/medibloc/panacea-core/v2 v2.1.0-alpha2.0.20220915052104-fd4a0ce6c4fb
+	github.com/medibloc/panacea-core/v2 v2.1.0-alpha2.0.20220916060638-9a0c908458a3
 	github.com/sirupsen/logrus v1.9.0
 	github.com/spf13/cobra v1.5.0
 	github.com/spf13/viper v1.12.0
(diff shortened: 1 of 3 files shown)


gyuguen:
I'll have to check more, but looking at  [this code](https://github.com/cosmos/cosmos-sdk/blob/v0.45.4/x/params/types/subspace.go#L108) it unmarshals using legacyAmino from cosmos-sdk.

This is a very unusual case of using legacyAmino instead of using protoBuf to insert data.

However, it is unclear whether this has a direct effect.

youngjoon-lee:
> I'll have to check more, but looking at [this code](https://github.com/cosmos/cosmos-sdk/blob/v0.45.4/x/params/types/subspace.go#L108) it unmarshals using legacyAmino from cosmos-sdk.
> 
> This is a very unusual case of using legacyAmino instead of using protoBuf to insert data.
> 
> However, it is unclear whether this has a direct effect.

Yeah. I'm not sure. Let's merge the PR https://github.com/medibloc/panacea-core/pull/407, then we're gonna be happy :)

gyuguen (approved):
Good

inchori (approved):
lgtm

youngjoon-lee:
> Yeah. I'm not sure. Let's merge the PR [medibloc/panacea-core#407](https://github.com/medibloc/panacea-core/pull/407), then we're gonna be happy :)

Just merged the https://github.com/medibloc/panacea-core/pull/407, and updated the dep in this PR: https://github.com/medibloc/panacea-doracle/pull/70/commits/65ecbfe118f26750f6b0ebc345e5f8a582d617fe

PR #72, opened 2022-09-16
ci: separate the Lint job with the main job

youngjoon-lee:
Because golangci-lint eats lots of memory, I'd like to separate the Lint job with the main job. The main job will be run on the self-hosted runner as it is. The Lint job will be run on the Github-hosted runner. It will reduce the cost of our self-hosted machine.

FYI, I used the latest version of golangci-lint which fixes some unexpected errors. For that, I had to bump the version of `actions/setup-go` and `actions/checkout`. 

FYI, I will rebase my PR #56 to this PR.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,13 +13,13 @@ jobs:
 
     steps:
       - name: Set up Go 1.x
-        uses: actions/setup-go@v2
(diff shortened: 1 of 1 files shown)


0xHansLee (approved):
goood

gyuguen (approved):
Thanks

PR #73, opened 2022-09-16
feat: set log level for tmlog (for light client)

youngjoon-lee:
By default, tmlog doesn't filter logs according to the log level. So, we should enable the filter.


audtlr24 (approved):
Thanks, lgtm!

0xHansLee (approved):
lgtm

PR #76, opened 2022-09-19
fix: use SHA256 as KDF to derive a shared key from a ECDH shared secret

### LINKED ISSUE #75: Invalid shared key is generated very rarely
We use ECDH to generate a shared key for symmetric encryption, specifically AES256 (which requires a 256-bit key).
However, the [SharedKey(priv, pub)](https://github.com/medibloc/panacea-doracle/blob/5cee3436383cf1c3c43cb624de44a8c5d0d80dc8/crypto/key.go#L32) returns a non-256-bit key very rarely.

CI failure:  https://github.com/medibloc/panacea-doracle/actions/runs/3079307089/jobs/4975449849

My reproduction with adding some logs: 
```
=== RUN   TestDecryptWithAES256
    encrypt_test.go:80: len(privKey1):32, len(pubKey1):65
    encrypt_test.go:83: len(privKey2):32, len(pubKey2):65
    encrypt_test.go:89: len(sharedKey1):31, len(sharedKey2):31
    encrypt_test.go:96:
        	Error Trace:	/Users/yjlee/repos/panacea-doracle/crypto/encrypt_test.go:96
        	Error:      	Received unexpected error:
        	            	secret key is not for AES-256: total 248 bits
        	Test:       	TestDecryptWithAES256
--- FAIL: TestDecryptWithAES256 (0.00s)
```

It is because a shared secret generated by ECDH (`btcec.GenerateSharedSecret(priv, pub)`) is not always 256 bits. To use the shared secret as a cryptographic key, we should use a KDF (key derivation function) to derive a shared key from the shared secret.
There are a lot of KDFs that we can use, but I think we can simply use SHA256 in our cases.

References:
- https://github.com/decred/dcrd/blob/af2367a161a95b89eeedf304a870e463bfa798a3/dcrec/secp256k1/ecdh.go#L12
  - And, https://github.com/decred/dcrd/blob/af2367a161a95b89eeedf304a870e463bfa798a3/dcrec/secp256k1/example_test.go#L55
- https://stackoverflow.com/questions/35265635/key-derivation-function-for-shared-secret-generated-by-diffie-hellman-key-exchan

youngjoon-lee:
Closes #75


0xHansLee (approved):
LGTM!

audtlr24 (approved):
Thanks, lgtm!

inchori (approved):
LGTM 👍

PR #77, opened 2022-09-20
fix: add tls certificate in connection with https endpoint

### LINKED ISSUE #71: fix url scheme issue for gRPC endpoint
When we create gRPC connection to the devnet, we've got error of too many colons in address. 
Fix this issue to create connection with an endpoint of the devnet.

0xHansLee:
Close #71 

If the endpoint of gRPC is https, x509 certificate is added as credential for TLS connection.

FYI) Based on [RFC 3986](https://www.rfc-editor.org/rfc/rfc3986), `GrpcAddr` is to be received as absolute url including scheme to be parsed correctly using `url.Parse`.
Of course, we can add default scheme(e.g. "http") if not exist. But I think it would be better to use the full absolute url for all urls for consistency and to be clear.


panacea/grpc_client.go
```diff
@@ -16,7 +18,26 @@ type GrpcClient struct {
...
+	}
+
+	var cred grpc.DialOption
+
+	if parsedUrl.Scheme == "https" {
+		tlsConfig, err := config.CreateTLSConfig()
+		if err != nil {
+			return nil, fmt.Errorf("failed to create TLS cofig: %w", err)
+		}
+		cred = grpc.WithTransportCredentials(credentials.NewTLS(tlsConfig))
```
youngjoon-lee: IIRC, we can just use a zero `tls.Config{}`.
```go
package main

import (
	"context"
	"crypto/tls"
	"fmt"

	banktypes "github.com/cosmos/cosmos-sdk/x/bank/types"
	"google.golang.org/grpc"
	"google.golang.org/grpc/credentials"
)

func main() {
	addr := "grpc.gopanacea.org:443"

	creds := credentials.NewTLS(&tls.Config{})
	conn, err := grpc.DialContext(context.Background(), addr, grpc.WithTransportCredentials(creds))
	if err != nil {
		panic(err)
	}
	defer conn.Close()

	client := banktypes.NewQueryClient(conn)
	resp, err := client.TotalSupply(context.Background(), &banktypes.QueryTotalSupplyRequest{})
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v\n", resp)
}
```

0xHansLee: Oh, I didn't know that. I'll change it as you suggested. Thanks!

gyuguen (approved):
Thanks

youngjoon-lee (approved):
lgtm

PR #79, opened 2022-09-20
fix: Fixed to vote for invalid trusted block info

### LINKED ISSUE #78: In case of incorrect trusted block info in the Oracle registration verification process, voting is not performed.
When validating a new oracle, if invalid trusted block info exists, there is a problem that it terminates without voting.
I improve to vote no in case of invalid trusted block info.

gyuguen:
Closes #78

### DIFF
--- a/event/register_oracle_event.go
+++ b/event/register_oracle_event.go
@@ -16,6 +16,7 @@ import (
 	"github.com/medibloc/panacea-doracle/panacea"
 	"github.com/medibloc/panacea-doracle/sgx"
 	log "github.com/sirupsen/logrus"
+	"github.com/tendermint/tendermint/light/provider"
 	ctypes "github.com/tendermint/tendermint/rpc/core/types"
 )
 
@@ -60,7 +61,10 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
 		return err
 	}
 
-	voteOption := verifyAndGetVoteOption(oracleRegistration, e)
+	voteOption, err := verifyAndGetVoteOption(oracleRegistration, e)
+	if err != nil {
+		return err
+	}
 
 	msgVoteOracleRegistration, err := makeOracleRegistrationVote(uniqueID, e.reactor.OracleAcc().GetAddress(), addressValue, voteOption, e.reactor.OraclePrivKey().Serialize(), oracleRegistration.NodePubKey, oracleRegistration.Nonce)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


gyuguen:
I wanted to add test code, but it wasn't easy.
So, the test was conducted manually.
It seems that refactoring is necessary for a smooth test.

event/register_oracle_event.go
```diff
@@ -89,8 +81,25 @@ func (e RegisterOracleEvent) EventHandler(event ctypes.ResultEvent) error {
...
-// verifyReportAndGetVoteOption validates the RemoteReport and returns the voting result according to the verification result.
-func verifyReportAndGetVoteOption(oracleRegistration *types.OracleRegistration, e RegisterOracleEvent) types.VoteOption {
+// verifyAndGetVoteOption performs a verification to determine a vote.
+// - Verify that trustedBlockInfo registered in OracleRegistration is valid
+// - Verify that the RemoteReport is valid
+func verifyAndGetVoteOption(oracleRegistration *types.OracleRegistration, e RegisterOracleEvent) types.VoteOption {
+	block, err := e.reactor.QueryClient().GetLightBlock(oracleRegistration.TrustedBlockHeight)
+	if err != nil {
+		log.Warnf("failed to get lightBlock. height(%v), error(%v)", oracleRegistration.TrustedBlockHeight, err)
+		return types.VOTE_OPTION_NO
```
0xHansLee: If an error occurs when getting light block, I think we should return error and not vote for no.
In this case, something's wrong in oracle itself and this oracle cannot judge the validity of oracle registration.

gyuguen: I first developed it as @H4NLee said.
However, if the height of the parameter is higher than the current block, an error occurs in this process.

Umm, I think it would be better to vote no for certain errors and not vote for the rest.

0xHansLee: Ah got it. Then, maybe [`ErrHeightTooHigh`](https://github.com/tendermint/tendermint/blob/c69ab6884837ad7a9e6ebadc2598d702bd14f5d5/light/provider/errors.go#L11)?

gyuguen: Right!

gyuguen: I fixed it. (https://github.com/medibloc/panacea-doracle/pull/79/commits/f082261652308262a129849d39f7976d72635afa)

0xHansLee (approved):
LGTM

inchori (approved):
LGTM, Thanks!

PR #88, opened 2022-10-19
feat: add an IPFS storage configuration

inchori:
Implement an IPFS storage configuration.
- IPFS's node address configuration was set in `config.go`, also in `service.go`.
- The `ipfs.go` which has a `Add` and `Get` method was implemented in `store` package.
- Add an `json_schema.go` refers to [panacea-coracle function](https://github.com/medibloc/panacea-coracle/blob/deee95ff5a1ed3bea935c45f5a81826c2b05bf0b/validation/jsonschema.go).
- Use [go-ipfs-api](https://github.com/ipfs/go-ipfs-api).


gyuguen:
Good.
How about adding  test code that run ipfs?

inchori:
@gyuguen. Ok I'll add it and re-request. For now, I'll draft this branch for run ipfs test.

inchori:
The `ipfs_test.go` will be implemented in another branch with docker test which is running IPFS.

store/ipfs.go
```diff
@@ -0,0 +1,62 @@
+package store
```
0xHansLee: How about making it a separate package `ipfs`?

store/ipfs.go
```diff
@@ -0,0 +1,62 @@
...
+	newStr := buf.String()
+
+	var deal types.Deal
+
+	err = json.Unmarshal([]byte(newStr), &deal)
+	if err != nil {
+		return nil, err
+	}
+
+	return deal.DataSchema, nil
```
0xHansLee: To make this `Get` function more general, I think that this function can simply return `buf.Bytes()`.

store/testdata/test_deal.json
```diff
@@ -0,0 +1,16 @@
+{
```
0xHansLee: Where this test data will be used for?

config/config.go
```diff
@@ -38,6 +40,10 @@ type PanaceaConfig struct {
 	LightClientWitnessAddrs []string `mapstructure:"light-client-witness-addrs"`
 }
 
+type IpfsConfig struct {
+	IpfsNodeAddr string `mapstructure:"ipfs-addr"`
```
0xHansLee:
```suggestion
	IpfsNodeAddr string `mapstructure:"ipfs-node-addr"`
```

config/toml.go
```diff
@@ -49,6 +49,14 @@ light-client-primary-addr = "{{ .Panacea.LightClientPrimaryAddr }}"
...
 
 light-client-witness-addrs= "{{ StringsJoin .Panacea.LightClientWitnessAddrs "," }}"
+
+###############################################################################
+###                         Ipfs Configuration                           ###
+###############################################################################
+
+[ipfs]
+
+ipfs-addr = "{{ .Ipfs.IpfsNodeAddr }}"
```
0xHansLee:
```suggestion
ipfs-node-addr = "{{ .Ipfs.IpfsNodeAddr }}"
```

store/testdata/test_deal.json
inchori: Actually, this file was used in `TestIpfsGet()` which is located in `ipfs_test.go`.
This is now useless, so I'll remove it.

inchori: Also, I'll delete `testdata` directory. This is now useless.

store/ipfs.go
inchori: I put the `ipfs` under the `store` package because I think that `ipfs`'s role is also used as `store`. But in future, some file or method that is associated with `ipfs` will be generated. So I'll make `ipfs` package.

store/ipfs.go
inchori: Is it returns the bytes of data and it will be unmarshalled in when verifying data?

config/toml.go
inchori: Fixed in [e22cb5a](https://github.com/medibloc/panacea-doracle/pull/88/commits/e22cb5ac922d2ee9ceb2e1be692480ae83f0edda).

youngjoon-lee (approved):
lgtm

audtlr24 (approved):
Thanks, LGTM 👍

gyuguen (approved):
LGTM

store/ipfs.go
0xHansLee: Right. In `ipfs`, I think there is no need to unmarshal the data. Although it depends on the data the caller get, and its the job the caller should do :)

0xHansLee (approved):
lgtm

PR #89, opened 2022-10-20
feat: add DataVerificationVote event handler

### LINKED ISSUE #80: Implement a Data Verification Event Handler
Implement a Data Verification Event Handler.
- Subscribe a `SellData` transaction event.
- Get CID from panacea and get data with CID from IPFS or storage.
- Verifying a data and executing vote transaction.

inchori:
Closes #80 

Implement DataVerificationVote event handler
- Implement a Data Verification Event handler which subscribes `message.action='SellData'` event.
- Get a data with `verifiableCID` that is gotten from `dataSale` in panacea.
- Verify the json schema of data which is from `IPFS`.
- Compare the `dataHash` in `IPFS` with `dataHash` in panacea.
- The `data_verification_event_test.go` will be implemented when the docker test is configured.


event/datadeal/data_verification_event.go
```diff
@@ -1 +1,215 @@
...
+	oraclePrivKey := d.reactor.OraclePrivKey()
+
+	decryptedData, err := d.decryptData(oraclePrivKey, dataSale, deal, encryptedDataBz)
+	if err != nil {
+		return err
+	}
+
+	if !d.compareDataHash(dataSale, decryptedData) {
+		return fmt.Errorf("invalid data hash (%s)", dataSale.DataHash)
+	}
```
inchori: I have a question. If the data hash comparison is failed, should the oracle vote NO or just show a warn message like this?

audtlr24: IMO, it seems good to vote NO even in this case,.

inchori:
After manual test, I'll re-open this PR.

event/datadeal/data_verification_event.go
inchori: Fixed in [fb78115](https://github.com/medibloc/panacea-doracle/pull/89/commits/fb78115e52ab125bc5e3c7bc57d68128e258309e) and [ce7fa24](https://github.com/medibloc/panacea-doracle/pull/89/commits/ce7fa248bf552b40bedbd15b7330d9f440879ff3).

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,215 @@
...
+
+	if err := broadcastTx(d.reactor.GRPCClient(), txBytes); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (d DataVerificationEvent) decryptData(oraclePrivKey *btcec.PrivateKey, dataSale *types.DataSale, deal *types.Deal, encryptedDataBz []byte) ([]byte, error) {
```
0xHansLee: Currently, this function do two things: getting account and decrypting data.
I suggest to limit this function to do decrypt data because its name is `decryptData`.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,215 @@
...
+	msgVoteDataVerification := &types.MsgVoteDataVerification{
+		DataVerificationVote: dataVerificationVote,
+		Signature:            sig,
+	}
+
+	return msgVoteDataVerification, nil
+}
+
+// generateTxBytes generates transaction byte array.
+// TODO: generateTxBytes function will be refactored.
```
0xHansLee, line 197: Are you considering to make this function reusable?

event/datadeal/data_verification_event.go
0xHansLee: IMO, it should return `NO`. Because it means that the seller try to fake data which is different from the one in panacea.

event/datadeal/data_verification_event.go
inchori: Okay, I'll separate it!

event/datadeal/data_verification_event.go
inchori: Yes, this function used in every events. So I'll refactor this function reusable.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+}
+
+func (d DataVerificationEvent) verifyDataSaleAndGetVoteOption(dataSale *types.DataSale, decryptedData []byte, dataSchema []string) oracletypes.VoteOption {
+	if !d.compareDataHash(dataSale, decryptedData) {
+		return oracletypes.VOTE_OPTION_NO
+	}
+
+	err := validation.ValidateJSONSchemata(decryptedData, dataSchema)
+	if err != nil {
+		log.Warnf("failed to verify data. error(%s)", err)
```
gyuguen: I think the log type would be `info`.
Because this is the correct check.
When I got a `warn` log type, I will think this is bad process.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+
+func (d DataVerificationEvent) GetEventAttributeKey() string {
+	return "action"
+}
+
+func (d DataVerificationEvent) GetEventAttributeValue() string {
+	return "'SellData'"
+}
+
+func (d DataVerificationEvent) EventHandler(event ctypes.ResultEvent) error {
```
gyuguen, line 47: I think oracle should be vote if there are no special problem.
For example, if oracle want to get deal, but deal is not exist, then oracle shoud be voted `no`.
But if oracle want to get deal, but received another error(exclude the not exist deal error), oracle should not vote.

gyuguen (commented):
Thank you.
Please check my comment :)

event/datadeal/data_verification_event.go
0xHansLee: It makes sense. 
We have to make `EventHandler` to vote by judging whether the data is correct or not.

> For example, if oracle want to get deal, but deal is not exist, then oracle shoud be voted no.

The former case gyuguen mentioned is definitely invalid data sale. No one can sell their data to non-existing deal.

gyuguen: Right. 
I think oracle cannot verify data if not eixst deal. 
So oracle votes no.

Does @H4NLee think when this case oracle should not voted?

0xHansLee: I meant that if there is no deal (when getting deal, got `NotExist` err), oracle should vote `NO`. 
The data sale for selling data on not existing deal should also be considered invalid I think.

gyuguen: Aha, I understood your answer.
Thank you

0xHansLee: In addition to the case gyuguen said above, let me take a look at other cases.

inchori: So the if any errors emerged, the oracle votes `NO`. Am I right?

0xHansLee: Maybe not. 

> For example, if oracle want to get deal, but deal is not exist, then oracle shoud be voted no.
But if oracle want to get deal, but received another error(exclude the not exist deal error), oracle should not vote.

As gyuguen said, we need to decide to vote or not depending on the errors. That's why I'm taking look at the cases.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+
+	dealID, err := strconv.ParseUint(dealIDStr, 10, 64)
+	if err != nil {
+		return err
+	}
+
+	deal, err := d.reactor.QueryClient().GetDeal(dealID)
+	if err != nil {
+		return err
+	}
```
0xHansLee: If error is `NotFound`, vote `NO`.
Otherwise, return error and do nothing.

I think it's impossible to sell the data to non-existing deal, but it make sense to vote `NO` with non-existing deal.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+
+	dataSale, err := d.reactor.QueryClient().GetDataSale(dataHash, dealID)
+	if err != nil {
+		return err
+	}
+
+	encryptedDataBz, err := d.reactor.Ipfs().Get(dataSale.VerifiableCid)
+	if err != nil {
+		return err
+	}
```
0xHansLee: IMO, if an error occurs when getting data from ipfs, oracle should vote `NO` whatever the reason is.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+
+	sellerPubKeyBytes := sellerAcc.GetPubKey().Bytes()
+
+	sellerPubKey, err := btcec.ParsePubKey(sellerPubKeyBytes, btcec.S256())
+	if err != nil {
+		return err
+	}
```
0xHansLee: When seller address is invalid, should oracle vote `NO`?
I'm not sure :(

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,216 @@
...
+
+	decryptSharedKey := crypto.DeriveSharedKey(oraclePrivKey, sellerPubKey, crypto.KDFSHA256)
+
+	decryptedData, err := d.decryptData(decryptSharedKey, deal.Nonce, encryptedDataBz)
+	if err != nil {
+		return err
+	}
```
0xHansLee: If decryption fails, doracle may vote `NO` as well. I think it's invalid data sale.

0xHansLee (commented):
Based on @gyuguen's [comment](https://github.com/medibloc/panacea-doracle/pull/89#discussion_r1006309293), I left some comments regarding error handling.

event/datadeal/data_verification_event.go
inchori: Thanks for comment. I have a question. How about change a return type of `EventHandler` method of `Event` interface to `types.VOTE_OPTION`? So it can be also changed in `register_oracle_event`.

inchori:
I'll also reflect this [comment](https://github.com/medibloc/panacea-doracle/pull/93#issuecomment-1294726609) to this branch.

inchori:
I reflected the comments in [0ee46f8](https://github.com/medibloc/panacea-doracle/pull/89/commits/0ee46f810c699dadbcd9b80f86644dfdaaad616b).

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,231 @@
...
+func (d DataVerificationEvent) EventHandler(event ctypes.ResultEvent) error {
+	dealIDStr := event.Events[types.EventTypeDataVerificationVote+"."+types.AttributeKeyDealID][0]
+	dataHash := event.Events[types.EventTypeDataVerificationVote+"."+types.AttributeKeyDataHash][0]
+
+	dealID, err := strconv.ParseUint(dealIDStr, 10, 64)
+	if err != nil {
+		return err
+	}
+
+	voteOption := d.verifyAndGetVoteOption(dealID, dataHash)
```
audtlr24: If the voteOption is `VOTE_OPTION_UNSPECIFIED', it will generate meaningless vote tx. At this time, oracle do not vote.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,231 @@
...
+	decryptedData, err := d.decryptData(decryptSharedKey, deal.Nonce, encryptedDataBz)
+	if err != nil {
+		return nil, err
+	}
+
+	return decryptedData, nil
+}
+
+func (d DataVerificationEvent) verifyAndGetVoteOption(dealID uint64, dataHash string) oracletypes.VoteOption {
+	deal, err := d.reactor.QueryClient().GetDeal(dealID)
```
audtlr24, line 133: I think the internal error message(`err`) is gone. The below dataSale also print internal error as well.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,231 @@
...
+	if dataSale.Status != types.DATA_SALE_STATUS_VERIFICATION_VOTING_PERIOD {
+		log.Info("dataSale's status is not DATA_SALE_STATUS_VERIFICATION_VOTING_PERIOD")
+		return oracletypes.VOTE_OPTION_UNSPECIFIED
+	}
+
+	decryptedData, err := d.convertSellerData(deal, dataSale)
+	if err != nil {
+		log.Infof("failed to decrypt seller data, error (%s)", err)
+		return oracletypes.VOTE_OPTION_NO
+	}
```
0xHansLee: IMHO, doracle should vote for `NO` only when the data decryption fails.

audtlr24 (commented):
Thanks, I have some comments.

event/datadeal/data_verification_event.go
0xHansLee: agree

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,231 @@
...
+
+	dataSale, err := d.reactor.QueryClient().GetDataSale(dataHash, dealID)
+	if err != nil {
+		if errors.Is(err, types.ErrDataSaleNotFound) {
+			log.Infof("failed to find dataSale (%s)", dataHash)
+			return oracletypes.VOTE_OPTION_NO
+		} else {
+			return oracletypes.VOTE_OPTION_UNSPECIFIED
+		}
+	}
```
0xHansLee: `ErrDataSaleNotFound` means no data sale for vote ([as gyuguen said](https://github.com/medibloc/panacea-doracle/pull/91#discussion_r1009156876)).
That is, vote tx is meaningless, so we can return `oracletypes.VOTE_OPTION_UNSPECIFIED` for all errors in query of data sale.

inchori: Fixed in [8ee1242](https://github.com/medibloc/panacea-doracle/pull/89/commits/8ee1242b70252bc579b95b0e16ccbe204ba77e01).

event/datadeal/data_verification_event.go
inchori: Fixed in [8ee1242](https://github.com/medibloc/panacea-doracle/pull/89/commits/8ee1242b70252bc579b95b0e16ccbe204ba77e01).

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,231 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return decryptedData, nil
+}
+
+func (d DataVerificationEvent) verifyAndGetVoteOption(dealID uint64, dataHash string) (oracletypes.VoteOption, error) {
+	deal, err := d.reactor.QueryClient().GetDeal(dealID)
+	if err != nil {
```
audtlr24, line 134: We decided to vote `VOTE_OPTION_NO` for a `NotFound` error during `GetDeal` or `GetDataSale` query. This is because it is difficult to see this situation as `oracle`'s fault.

audtlr24 (commented):
Thanks 👍 , I have one comment.

event/datadeal/data_verification_event.go
audtlr24: Sorry. Voting for `dataSale` that does not exist is meaningless, so it seems to be okay.

audtlr24 (approved):
LGTM!

event/datadeal/data_verification_event.go
gyuguen: Umm, I think this case shoud be voted `NO`.
If the merchant address is invalid, it could be the wrong logic in the blockchain code.

So, oracle blocks this vote from passing.

event/datadeal/data_verification_event.go
```diff
@@ -1 +1,232 @@
...
+		return oracletypes.VOTE_OPTION_UNSPECIFIED, errors.New("dataSale's status is not DATA_SALE_STATUS_VERIFICATION_VOTING_PERIOD")
+	}
+
+	decryptedData, err := d.convertSellerData(deal, dataSale)
+	if err != nil {
+		log.Infof("failed to decrypt seller data, error (%v)", err)
+		return oracletypes.VOTE_OPTION_NO, err
+	}
+
+	if !d.compareDataHash(dataSale, decryptedData) {
```
gyuguen, line 152: if oracle failed get seller's data, this part occur error and it will be vote `NO`. Did I understand correctly?

gyuguen (commented):
Thank you.
I left one question

gyuguen (approved):
And LGTM

event/datadeal/data_verification_event.go
inchori: Did you mean the `compareDataHash` method? or the line 158?

gyuguen: Yes, I have a question about `compareDataHash` function

gyuguen: And based function `convertSellerData`

gyuguen: detail lines: https://github.com/medibloc/panacea-doracle/pull/89/files#diff-77b26936f5a1a147ce3878a5f0945af04ae92defbf19d41f4ef201713de9e9e9R109

event/datadeal/data_verification_event.go
0xHansLee: OK. it makes sense.
Oracle validates all things and if something's wrong, vote `NO`.

event/datadeal/data_verification_event.go
inchori: The `compareDataHash` occur errors when after get seller's data and `convertSellerData`.

event/datadeal/data_verification_event.go
inchori: Fixed in [aa6011f](https://github.com/medibloc/panacea-doracle/pull/89/commits/aa6011ffc81d7fe8369ab22544962903c7bc7d85).

event/datadeal/data_verification_event.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return decryptedData, nil
+}
+
+func (d DataVerificationEvent) verifyAndGetVoteOption(dealID uint64, dataHash string) (oracletypes.VoteOption, error) {
+	deal, err := d.reactor.QueryClient().GetDeal(dealID)
+	if err != nil {
+		log.Infof("failed to find deal (%d)", dealID)
+		return oracletypes.VOTE_OPTION_NO, nil
+	}
```
0xHansLee: Shouldn't we handle errors differently here?
IMO, oracle should vote `NO` only in case of `datadealtypes.ErrDealNotFound`. 
In other cases, oracle should not vote (`oracletypes.VOTE_OPTION_UNSPECIFIED`).

event/datadeal/data_verification_event.go
```diff
@@ -0,0 +1,230 @@
...
+	decryptedDataHashStr := hex.EncodeToString(decryptedDataHash[:])
+
+	return decryptedDataHashStr == dataSale.DataHash
+}
+
+func (d DataVerificationEvent) convertSellerData(deal *types.Deal, dataSale *types.DataSale) ([]byte, error) {
+	encryptedDataBz, err := d.reactor.Ipfs().Get(dataSale.VerifiableCid)
+	if err != nil {
+		log.Infof("failed to get data from IPFS: %v", err)
+		return nil, nil
```
0xHansLee: maybe?
```suggestion
		return nil, err
```

event/datadeal/data_verification_event.go
inchori: Yes I agree. So I need to change like [this code](https://github.com/medibloc/panacea-doracle/blob/18d2e2c1e481a2201d101f6724238621805cbee4/event/datadeal/data_delivery_event.go#L111).
Is it right?

0xHansLee: Yes, that's what I meant. 👍

inchori: Thanks! Fixed in [1be1c83](https://github.com/medibloc/panacea-doracle/pull/89/commits/1be1c830c1df3159a0e1bad2e63464ed202ef213).

inchori:
I changed `VOTE_OPTION_UNSPECIFIED` to` VOTE_OPTION_NO` when every errors are occured in [0717390](https://github.com/medibloc/panacea-doracle/pull/89/commits/07173907d5c87b1d02e6d74a6c9e454a16084a81).

PR #90, opened 2022-10-20
feat: implement request an oracle upgrade

### LINKED ISSUE #85: Implements the ability to migrate OraclePrivKey to a new Oracle's binary.
## Background
If the Oracle binary is changed, the previously sealed `OraclePrivKey` cannot be used.
This is because only binaries signed under the same conditions can be unsealed.
Therefore, a transaction must be performed in [panacea-core](https://github.com/medibloc/panacea-core) so that `OraclePrivKey` can be delivered to the new binary.

## Implements
UpgradeOracle transaction must be performed with panacea-core. 
The command to be executed is as follows.
```
doracled upgrade-oracle --trusted-block-hash {hash} --trusted-block-height {height}
```

Before performing a transaction, the following process is followed.
* Create a NodePrivKey.
* Creates QueryClient with trustedBlockInfo received as flag.
* Get the address with MNEMONIC set in config.
* Get the UniqueID.
* Create a nonce

Execute a transaction in panacea-core with this generated value.

gyuguen:
Closes #85 

## Implement
* Added oracle upgrade request cmd


cmd/doracled/cmd/upgrade_oracle.go
```diff
@@ -0,0 +1,116 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			conf, err := loadConfigFromHome(cmd)
+			if err != nil {
+				return err
+			}
+			ctx := context.Background()
+
+			// if node key exists, return error.
+			nodePrivKeyPath := conf.AbsNodePrivKeyPath()
+			if os.FileExists(nodePrivKeyPath) {
```
0xHansLee, line 34: In general, there is an existing node key for oracle registration (or for previous version).
That is, most user would encounter this error. Is it ok? Or overwrite the existing node key when upgrading?
One more option is to use separate dir for node key depending on uniqueID as you suggested before.

gyuguen: At upgrade time, the HOME path will probably be set differently.
Because we can't run multi doracles in the same path.
So, this part uses the existing `ReigsterOracle` function as it is.

0xHansLee: Right. I missed that point. Thx

0xHansLee (approved):
lgtm

audtlr24 (approved):
Thanks, lgtm!
There is typo in PR message 😄

PR #91, opened 2022-10-21
feat: implement oracle upgrade handling

### LINKED ISSUE #86: Implement OracleUpgradeEvent handling for Oracle upgrade support.
## Background
The new binary oracle wants to get the `OraclePrivKey` stored by the old binary oracle.
In order to pass this `OraclePrivKey`, old binary oracles must verify that the new binary oracle is really valid.
Also, if valid, the old binary oracle must vote for **panacea-core** by encrypting the `OraclePrivKey` so that the new binary oracle can decrypt it.

## Implement
Old binary oracles should detect the events below.
```
message.action=OracleUpgrade
```

If an event is detected, the following process is performed.
* Query  Oracle information to upgrade from `OracleRegistration`
* Verify status a OracleRegistration
  * Only `VOTING_PERIOD` is valid
* Verify that `OracleUpgradeProposal` passed
  * If upgrade information exists in `OracleUpgradeInfo`, it means that it passed.
* Compare if `UniqueID` in `OracleUpgradeInfo` and `OracleRegistration` are the same.
* Verify the `TrustedBlockHash` and `TrustedBlockHeight` obtained from `OracleRegistration`.
* Verify the `NodePubKeyRemoteReport` obtained from `OracleRegistration`.
  * Basic RemoteREport validation
  * Verify hashed `NodePubKey` stored in RemoteReport
* If all verification passes, perform `OracleRegistrationVote` transaction in **panacea-core**.

gyuguen:
Closes #86 

## Implement
* Detect OracleUpgradeEvent to verify upgrade status.
* Vote Yes or No according to the verification result.


event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,94 @@
...
+	upgradeInfo, err := e.service.QueryClient().GetOracleUpgradeInfo()
+	if err != nil {
+		if errors.Is(err, panacea.ErrEmptyValue) {
+			log.Infof("not exist oracle upgrade info")
+			return oracletypes.VOTE_OPTION_NO, nil
+		}
+		log.Errorf("failed to get oracle upgrade info. %v", err)
+		return oracletypes.VOTE_OPTION_NO, nil
+	}
+	if upgradeInfo.UniqueId != e.service.EnclaveInfo().UniqueIDHex() {
```
0xHansLee: IIRC, this handler would be executed in the old version of `doracle`.
That is, `e.service.EnclaveInfo()` might be the uniqueID of the old version of `doracle`.
Instead, we may have to compare the uniqueIDs from `OracleRegistration` and `upgradeInfo` if the requested uniqueID is equal to the target version of `doracle`

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,94 @@
...
+	return "'UpgradeOracle'"
+}
+
+func (e UpgradeOracleEvent) EventHandler(event ctypes.ResultEvent) error {
+	addressValue := event.Events[oracletypes.EventTypeUpgradeVote+"."+oracletypes.AttributeKeyOracleAddress][0]
+	service := e.service
+	queryClient := service.QueryClient()
+
+	uniqueID := service.EnclaveInfo().UniqueIDHex()
+	oracleRegistration, err := queryClient.GetOracleRegistration(addressValue, uniqueID)
```
0xHansLee: And If my previous comment is right, we need to query the oracle registration using the uniqueID of upgrade info, not the previous version of `doracle`.

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (e UpgradeOracleEvent) verifyAndGetVoteOption(r *oracletypes.OracleRegistration) (oracletypes.VoteOption, error) {
+	upgradeInfo, err := e.service.QueryClient().GetOracleUpgradeInfo()
+	if err != nil {
+		if errors.Is(err, panacea.ErrEmptyValue) {
+			log.Infof("not exist oracle upgrade info")
+			return oracletypes.VOTE_OPTION_NO, nil
+		}
+		log.Errorf("failed to get oracle upgrade info. %v", err)
+		return oracletypes.VOTE_OPTION_NO, nil
```
0xHansLee: If the error is `panacea.ErrEmptyValue`, I think this should return `NO` as is.
But in case of other error, how about returning `oracletypes.VOTE_OPTION_UNSPECIFIED`? or do not vote because there is an unexpected error while getting oracle upgrade info.

event/oracle/upgrade_oracle_event.go
gyuguen: Oh my mistake. You're right.
I haven't been able to test this after development.
I will correct this part. 
Thanks.

event/oracle/upgrade_oracle_event.go
gyuguen: Well on the blockchain there is no handling of `oracletypes.VOTE_OPTION_UNSPECIFIED`. 
Once this vote is received, we cannot tally the votes.
As @H4NLee said, it seems right not to vote if an unexpected error occurs.

event/oracle/expected.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/medibloc/panacea-doracle/config"
+	"github.com/medibloc/panacea-doracle/panacea"
+	"github.com/medibloc/panacea-doracle/sgx"
+)
+
+// reactor contains all ingredients needed for handling this type of event
+type reactor interface {
```
0xHansLee: why don't we share this `reactor` interface for all events?
Because a single service would be given to all events.

gyuguen: I didn't understand the question correctly.
This reactor is declared as a variable in every event.

0xHansLee: It is in package `oracle`. I meant to make this interface to be shared to other types of events, for example data verification. 
There is a similar interface defined in #80 as well (https://github.com/medibloc/panacea-doracle/blob/b2fa22dc5d5de2bfd5b394ffab29459ea74420ac/event/datadeal/data_verification_event.go#L32).

gyuguen: Oh, I understand.
The event was declared in datadeal as well.
I will refactor.
Thanks. :)

event/oracle/upgrade_oracle_event.go
0xHansLee: Good 👍. Let's just log the unexpected error then.

event/expected.go
```diff
@@ -0,0 +1,18 @@
+package event
```
gyuguen: I moved this interface to parent.

event/oracle/register_oracle_event.go
```diff
@@ -114,35 +104,38 @@ func (e RegisterOracleEvent) verifyAndGetVoteOption(oracleRegistration *types.Or
...
-func makeOracleRegistrationVote(uniqueID, voterAddr, votingTargetAddr string, voteOption types.VoteOption, oraclePrivKey, nodePubKey, nonce []byte) (*types.MsgVoteOracleRegistration, error) {
+func makeOracleRegistrationVote(uniqueID, voterAddr, votingTargetAddr string, voteOption oracletypes.VoteOption, oraclePrivKey, nodePubKey, nonce []byte) (*oracletypes.MsgVoteOracleRegistration, error) {
 	privKey, _ := crypto.PrivKeyFromBytes(oraclePrivKey)
 
-	pubKey, err := btcec.ParsePubKey(nodePubKey, btcec.S256())
-	if err != nil {
-		return nil, err
-	}
+	var encryptedOraclePrivKey []byte
+	if oracletypes.VOTE_OPTION_YES == voteOption {
```
gyuguen: If oracle voted a no, encryptedPrivKeys must be null.

event/oracle/register_oracle_event_no_docker_test.go
```diff
@@ -0,0 +1,52 @@
+package oracle
```
gyuguen, line 1: This test code does not use docker.

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,95 @@
...
+func (e UpgradeOracleEvent) GetEventAttributeValue() string {
+	return "'UpgradeOracle'"
+}
+
+func (e UpgradeOracleEvent) EventHandler(event ctypes.ResultEvent) error {
+	uniqueID := event.Events[oracletypes.EventTypeUpgradeVote+"."+oracletypes.AttributeKeyUniqueID][0]
+	addressValue := event.Events[oracletypes.EventTypeUpgradeVote+"."+oracletypes.AttributeKeyOracleAddress][0]
+	queryClient := e.reactor.QueryClient()
+
+	oracleRegistration, err := queryClient.GetOracleRegistration(addressValue, uniqueID)
```
gyuguen: If oracleRegistration is empty, it seems to wrong request.
So oracle does not vote this request.

event/oracle/upgrade_oracle_event.go
gyuguen: I fixed it!

gyuguen:
This PR changing status to DRAFT. Because oracle does failed validation to oracle upgrade.

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		log.Infof("failed to get oracleRegistration, voting ignored. uniqueID(%s), address(%s). %v", uniqueID, addressValue, err)
+		return err
+	}
```
0xHansLee: As done in #93 ([line](https://github.com/medibloc/panacea-doracle/pull/93/files#diff-fafe7cdb94cfba6f519545f3146bd5016f4ef6217768f2f97ae12355bda79d46R92)), error might be handled differently depending on the error is from doracle itself or not.
If the error is `NotFound`, doracle should vote `NO` I think.

gyuguen: Well, if oracleRegistration is not found, we cannot vote it.
Because we vote for oracleRegistration.

If we can't find oracleRegistration we can't vote because we don't have anyone to vote for. :)

gyuguen: like `dataSale`

0xHansLee: You're right. Thx for your works!

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		return oracletypes.VOTE_OPTION_UNSPECIFIED, fmt.Errorf("failed to get oracle upgrade info. %v", err)
+	}
+	if upgradeInfo.UniqueId != r.UniqueId {
+		return oracletypes.VOTE_OPTION_NO, fmt.Errorf("oracle's uniqueID does not match the uniqueID being upgraded. expected uniqueID(%s), oracle's uniqueID(%s), ",
+			upgradeInfo.UniqueId,
+			r.UniqueId)
+	}
+
+	block, err := e.reactor.QueryClient().GetLightBlock(r.TrustedBlockHeight)
```
audtlr24: Is there a reason for verifying the light block here? Is it already verified in queryClient?

audtlr24 (commented):
Thanks, I have one question.

event/oracle/upgrade_oracle_event.go
gyuguen: Your question is why validate block information in here?

gyuguen: If that's light,
When oracle's will be upgrade, it is changing uniqueID.
So the old lightClient data cannot used.

We used a new lightClient data in new oracle binary.

audtlr24: Ah.. sorry. There is light block info in `OracleRegistration`. So, we have to validate it... I had misunderstood.

audtlr24: Aha. There is another reason. Now I understand. thank you

audtlr24 (approved):
LGTM 👍

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,178 @@
...
+				uniqueID,
+				addressValue,
+				err)
+			return err
+		default:
+			log.Warnf("if an error occurs, no other voteOption is possible. uniqueID(%s), votingTargetAddress(%s) voteOption(%s), err(%v)",
+				uniqueID,
+				addressValue,
+				voteOption,
+				err)
```
0xHansLee: I don't think the case of `default` will happen, shouldn't an error be returned here?

0xHansLee (approved):
LGTM

event/oracle/upgrade_oracle_event.go
gyuguen: Current, we handler only `VOTE_OPTION_NO` and `VOTE_OPTION_UNSPECIFIED`.(Another code)
So I simply want printing log when unexpected set a voteOption.

Umm... Actually, I consider this part whether return error or not.
First of all, I will keep this code.

0xHansLee: Got it!

event/oracle/event.go
```diff
@@ -0,0 +1,103 @@
+package oracle
```
gyuguen, line 1: Functions commonly used by `OracleRegistrationVote` and `OracleUpgradeVote` are gathered here.

gyuguen (commented):
Please review again.
A lot of refactoring was done along with the content of the offline discussion.

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,173 @@
...
+
+	oracleRegistration, err := queryClient.GetOracleRegistration(votingTargetAddress, uniqueID)
+	if err != nil {
+		log.Infof("failed to get oracleRegistration. uniqueID(%s), address(%s). %v", uniqueID, votingTargetAddress, err)
+		return makeMsgVoteOracleRegistrationVoteTypeNo(uniqueID, voterAddress, votingTargetAddress, oraclePrivKeyBz)
+	}
+
+	voteOption, err := e.verifyAndGetVoteOption(oracleRegistration)
+	if err != nil {
+		log.Infof("vote No due to error while verify: %v", err)
```
0xHansLee, line 83: maybe missed the below
```go
return makeMsgVoteOracleRegistrationVoteTypeNo(uniqueID, voterAddress, votingTargetAddress, oraclePrivKeyBz)
```

event/oracle/upgrade_oracle_event.go
```diff
@@ -0,0 +1,173 @@
...
+		return fmt.Errorf("invalid product ID in the report")
+	}
+	if !bytes.Equal(report.SignerID, enclaveInfo.SignerID) {
+		return fmt.Errorf("invalid signer ID in the report")
+	}
+	uniqueID, err := hex.DecodeString(expectedUniqueID)
+	if err != nil {
+		return fmt.Errorf("invalid uniqueID format. uniqueID(%s). %w", uniqueID, err)
+	}
+	if !bytes.Equal(report.UniqueID, uniqueID) {
```
0xHansLee, line 147: Should we compare uniqueIDs from report and update info as well?
Maybe three uniqueIDs should be equal..?

event/oracle/upgrade_oracle_event.go
gyuguen: If voteOption equals `No`, `makeMsgVoteOracleRegistration` function execute voting `No`.
This line simply printings the log.

0xHansLee: I right.
As we discussed before, if error occurs voteOption might be `NO`.
No issue here. Thx!

gyuguen: ~~But it looks good to voting `No` this line.~~
~~I will fixed it.~~
For now, let's leave this code as it is.

event/oracle/upgrade_oracle_event.go
gyuguen: As I understood your comments, we already checked upgrade's uniqueID and oracleRegistration's uniqueID [here](https://github.com/medibloc/panacea-doracle/pull/91/files#diff-0c6d9bfa1427c39a9489d371dd13d4d5f847f7a562817ee976de39cefadf3bc1R104). :)

0xHansLee: I missed. Thank you.

0xHansLee (approved):
utACK

audtlr24 (approved):
I checked the refactored code. LGTM 👍

PR #93, opened 2022-10-28
feat: add DataDeliveryVote event handler

### LINKED ISSUE #81: Implement a Data Delivery Event Handler
Implement a Data Delivery Event Handler.
- Subscribe a `VoteDataVerification` end blocker event.
- Encrypt data by `buyerPubKey` and add it to IPFS or storage.
- The oracles executing vote transaction to new cid which is `delivery_cid`.
- If vote are passed, return the `delivery_cid` to panacea.

audtlr24:
Add DataDelieveryVote event handler
- this code pass the manual test on a normal `sellData` request
- But, there are two problems:
1) it takes a long time to process the requests (about 30sec), because there are 4 `QueryClient` requests and they are processed sequentially.
2) I don't know the exact reason, but the subscriber subscribe `DataDelivery event` of the `endBlocker` twice.

I will consider the above two issues and post them as another PR.


audtlr24:
Fixed to handle errors and votes correctly
- In most cases, we have voted through `VOTE_OPTION_NO` when it fails
- When `GetDeal` and `GetDataSale` get error that is not `NotFound` errors, do not vote and return an error

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,235 @@
...
+
+	if err := e.broadcastTx(e.reactor.GRPCClient(), txBytes); err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+func (e DataDeliveryVoteEvent) verifyAndGetVoteOption(dealID uint64, dataHash string, oraclePrivKey *btcec.PrivateKey) (oracletypes.VoteOption, string, error) {
```
gyuguen: I think it would be good write log when every case return from this function.
Because operators may want to know why the vote was decided.

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,235 @@
...
+	dealID, err := strconv.ParseUint(dealIDStr, 10, 64)
+	if err != nil {
+		return err
+	}
+
+	oraclePrivKey := e.reactor.OraclePrivKey()
+
+	voteOption, deliveredCid, err := e.verifyAndGetVoteOption(dealID, dataHash, oraclePrivKey)
+	if err != nil {
+		log.Infof("error while verify: %v", err)
```
gyuguen: I think if occur error, we print log as `error` type.
Of course, exclude `not found ...`

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,235 @@
...
+
+	return nil
+
+}
+
+func (e DataDeliveryVoteEvent) verifyAndGetVoteOption(dealID uint64, dataHash string, oraclePrivKey *btcec.PrivateKey) (oracletypes.VoteOption, string, error) {
+
+	dataSale, err := e.reactor.QueryClient().GetDataSale(dealID, dataHash)
+	if err != nil {
+		if err == datadealtypes.ErrDataSaleNotFound {
```
0xHansLee: For now, it seems no problem.
But if somehow the error `ErrDataSaleNotFound` is wrapped, this will not work. So I suggest to use `errors.Is()`.

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,235 @@
...
+	deliveredCID, err := e.makeDeliveredCid(deal, dataSale, oraclePrivKey)
+	if err != nil {
+		return oracletypes.VOTE_OPTION_NO, "", fmt.Errorf("error while make deliveredCid: %v", err)
+	}
+
+	return oracletypes.VOTE_OPTION_YES, deliveredCID, nil
+
+}
+
+func (e DataDeliveryVoteEvent) makeDeliveredCid(deal *datadealtypes.Deal, dataSale *datadealtypes.DataSale, oraclePrivKey *btcec.PrivateKey) (string, error) {
```
gyuguen: This function is to upload data that the buyer can decrypt to ipfs.
So I think it would be good to change function name to `convertBuyerDataAndUploadIpfs`
Well. Other names are possible, but I can't remember right now. :)

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,235 @@
...
+		if err == datadealtypes.ErrDataSaleNotFound {
+			return oracletypes.VOTE_OPTION_NO, "", err
+		} else {
+			return oracletypes.VOTE_OPTION_UNSPECIFIED, "", err
+		}
+	}
+
+	if dataSale.Status != datadealtypes.DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD {
+		return oracletypes.VOTE_OPTION_NO, "", nil
+	}
```
0xHansLee: IMO, if the voting for delivery is ended, oracle don't have to vote. wdyt?

panacea/query_client.go
```diff
@@ -395,3 +396,39 @@ func (q QueryClient) GetOracleParamsPublicKey() (*btcec.PublicKey, error) {
 
 	return btcec.ParsePubKey(pubKeyBz, btcec.S256())
 }
+
+func (q QueryClient) GetDeal(dealID uint64) (*datadealtypes.Deal, error) {
```
gyuguen, line 400: If possible, please add a test code. or if impossible don't added a test code :)
Maybe you can edited `/panacea/test/data/panacea-core.init.sh`

gyuguen (commented):
Thank you.
Please check my review

event/datadeal/data_delivery_event.go
0xHansLee: If this function is on `DataDeliveryVoteEvent` struct, you don't have to pass oracle private key.
The struct already has one :)

0xHansLee (commented):
Left some comments.

For improvements, we can cache the pubkey of seller and buyer (query for the first time, after first time reduce 2 queries)

And, if subscriber subscribes twice, it votes twice as well?

event/datadeal/data_delivery_event.go
gyuguen: I agree. If 'dataSale' is not able to vote, it is likely that Oracle will not have to vote.

gyuguen:
> Left some comments.
> 
> For improvements, we can cache the pubkey of seller and buyer (query for the first time, after first time reduce 2 queries)
> 
> And, if subscriber subscribes twice, it votes twice as well?

Yes. Good idea.
In general, values that do not change are cached at regular intervals.

event/datadeal/data_delivery_event.go
audtlr24: Then, the status is not `DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD`, oracle do not vote. right?

event/datadeal/data_delivery_event.go
audtlr24: Yes, I agree. But, oracle private key is used in `makeDataDeliveryVote`. So, I implemented it by passing the oracle private key obtained from the struct to avoid code duplication. Is it still inefficient to pass the oracle private key as a parameter?

audtlr24: Thanks, I fixed it. Please check again :)

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I change it to use `Errorf`.

event/datadeal/data_delivery_event.go
0xHansLee: right

event/datadeal/data_delivery_event.go
audtlr24: Okay, I change this function name to `convertBuyerDataAndAddToIpfs`, because `upload` is expressed as `add` in ipfs.

event/datadeal/data_delivery_event.go
0xHansLee: You mean `convertBuyerDataAndAddToIpfs`?
It is also a function on the struct, that is you can use the oracle private key by `e.reactor.OraclePrivKey()`

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I fixed it :)

event/datadeal/data_delivery_event.go
audtlr24: Thanks! Changed it to return `VOTE_OPTION_UNSPECIFIED`.

event/datadeal/data_delivery_event.go
audtlr24: Yes, I use `e.reactor.OraclePrivKey()`. And I pass it as parameter to `verifyAndGetVoteOption` and `makeDataDeliveryVote`. The `makeDataDeliveryVote` is function that make `msgVoteDataDelivery`.

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,236 @@
...
+	return nil
+
+}
+
+func (e DataDeliveryVoteEvent) verifyAndGetVoteOption(dealID uint64, dataHash string, oraclePrivKey *btcec.PrivateKey) (oracletypes.VoteOption, string, error) {
+
+	dataSale, err := e.reactor.QueryClient().GetDataSale(dealID, dataHash)
+	if err != nil {
+		if errors.Is(err, datadealtypes.ErrDataSaleNotFound) {
+			return oracletypes.VOTE_OPTION_NO, "", err
```
gyuguen: This case, we print `error` type log.
But I think if oracle does vote to no, we shouldn''t printing `error`  type log.

What I meant was to separate the logs into `info` types and `error` types.

audtlr24: Ah, I see. So I fixed it in [f95f622](https://github.com/medibloc/panacea-doracle/pull/93/commits/f95f622a0e79c9323ce203410674f9fce0f2e42c). Is is okay?

panacea/query_client.go
audtlr24: To make a `deal` in panacea, we need to put the `buyer_address` in the json file containing the deal info (e.g. create_deal.json) and generate a transaction with the json file as a input.
I tried modifying the `/panacea/test/data/panacea-core.init.sh`, but it doesn't work right away. I will try again and upload it through another PR.

event/datadeal/data_delivery_event.go
0xHansLee: I meant you don't have to pass the oracle private key.
Can just use the key from the `DataDeliveryVoteEvent` struct which has reactor and oracle private key.
For example,

```go
func (e DataDeliveryVoteEvent) EventHandler(event ctypes.ResultEvent) error {
    // ...
    voteOption, deliveredCid, err := e.verifyAndGetVoteOption(dealID, dataHash)
    // ...
}

func (e DataDeliveryVoteEvent) verifyAndGetVoteOption(dealID uint64, dataHash string) (oracletypes.VoteOption, string, error) {
    // ...
    deliveredCID, err := e.convertBuyerDataAndAddToIpfs(deal, dataSale, e.reactor.OraclePrivKey())
    // ...
}
```

event/datadeal/data_delivery_event.go
0xHansLee: Actually, when error occurs in `EventHandler`, doracle prints error log.
https://github.com/medibloc/panacea-doracle/blob/db3efc002fba6fae94742ae0fe455578c8f2318b/event/subscriber.go#L59

Thus, I suggest to log with the details of data sale, such as deal ID and data hash.

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I fixed it 👍.

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I remove the duplicated error log, and add detail info (dealID, dataHash) to error.

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (e DataDeliveryVoteEvent) GetEventAttributeKey() string {
+	return "vote_status"
+}
+
+func (e DataDeliveryVoteEvent) GetEventAttributeValue() string {
+	return "'started'"
+}
+
+func (e DataDeliveryVoteEvent) EventHandler(event ctypes.ResultEvent) error {
```
gyuguen, line 42: IMO, We need to print a log when execute transaction.
Because the way to know how and which requested was voted for can be found in Oracle's log.

event/datadeal/data_delivery_event.go
```diff
@@ -0,0 +1,231 @@
...
+		return oracletypes.VOTE_OPTION_UNSPECIFIED, "", errors.New("datasale status is not DATA_SALE_STATUS_DELIVERY_VOTING_PERIOD")
+	}
+
+	if len(dataSale.VerifiableCid) == 0 {
+		return oracletypes.VOTE_OPTION_NO, "", errors.New("there is no verifiableCid")
+	}
+
+	deal, err := e.reactor.QueryClient().GetDeal(dataSale.DealId)
+	if err != nil {
+		if errors.Is(err, datadealtypes.ErrDealNotFound) {
```
gyuguen: I think if does not exist deal, it does not return  `datadealtypes.ErrDealNotFound` error.
Maybe return to `"empty value"`. 
So check my code please.(https://github.com/medibloc/panacea-doracle/pull/91/files#diff-0c6d9bfa1427c39a9489d371dd13d4d5f847f7a562817ee976de39cefadf3bc1R110)

panacea/query_client.go
gyuguen: OK. It's not important now :)

gyuguen (commented):
Please check my comment :)

event/datadeal/data_delivery_event.go
audtlr24: I think it is necessary. But I'm not sure how to get the tx info from the event. Do you know how to get it?

gyuguen: Sorry, I wrote the wrong comment.
I think it would be good print before transaction. :)
So we can use msgVoteDataDelivery object.

if the transaction failed, we need to know the request information. (ref: https://github.com/medibloc/panacea-doracle/pull/91/files#diff-0c6d9bfa1427c39a9489d371dd13d4d5f847f7a562817ee976de39cefadf3bc1R87)

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I understand it. Fixed in [5246329](https://github.com/medibloc/panacea-doracle/pull/93/commits/524632931d95a3fd57d53d3ed561408287e664f1).

event/datadeal/data_delivery_event.go
audtlr24: Thanks, I add it in [ca895cd](https://github.com/medibloc/panacea-doracle/pull/93/commits/ca895cd008730615c9b1a6fc7889dbd574c43ab7).

gyuguen (approved):
LGTM

0xHansLee (approved):
LGTM 👍

PR #94, opened 2022-11-02
feat: change vote option when an error occurs

audtlr24:
- change `VOTE_OPTION_UNSPECIFIED` to `VOTE_OPTION_NO`
- upgrade go version to 1.19


event/datadeal/data_delivery_event.go
```diff
@@ -51,11 +51,7 @@ func (e DataDeliveryVoteEvent) EventHandler(event ctypes.ResultEvent) error {
 
 	voteOption, deliveredCid, err := e.verifyAndGetVoteOption(dealID, dataHash)
 	if err != nil {
-		if voteOption == oracletypes.VOTE_OPTION_UNSPECIFIED {
-			return fmt.Errorf("can't vote due to error while verify. dealID(%d). dataHash(%s): %v", dealID, dataHash, err)
-		} else {
-			log.Infof("vote NO due to error while verify. dealID(%d). dataHash(%s): %v", dealID, dataHash, err)
-		}
+		log.Infof("vote NO due to error while verify. dealID(%d). dataHash(%s): %v", dealID, dataHash, err)
```
inchori, line 54: I think that this part should be
```go
return fmt.Errorf("can't vote due to error while verify. dealID(%d). dataHash(%s)", dealID, dataHash)
 ```
 Wdyt?

audtlr24: Hmm.. I think we vote `VOTE_OPTION_NO`, not `can't vote`.
And, if it returns error, it can not make deliveryVote Tx. So, I think `log.Infof` is correct.

inchori: Ah yes, you're right 😂

0xHansLee (approved):
lgtm

inchori (approved):
LGTM

PR #95, opened 2022-11-03
feat: change event query function

audtlr24:
Change event query function to `GetEventQuery`.
- To solve the problem that Endblocker's event is subscribed twice, I change a query of `DataDeliveryEvent` to `tm.event='NewBlock' AND data_delivery.vote_status='started'`.
- So, I change function that get event query from event.

Minor Change
- In addition, a minor change is included in the `DataVerifiactionVote` part.
- upgrade cosmos-sdk version to `github.com/medibloc/cosmos-sdk v0.45.9-panacea.1`


event/datadeal/data_verification_event.go
```diff
@@ -53,61 +45,61 @@ func (d DataVerificationEvent) EventHandler(event ctypes.ResultEvent) error {
...
 	)
 	if err != nil {
 		return err
 	}
 
-	txBuilder := panacea.NewTxBuilder(*d.reactor.QueryClient())
+	txBuilder := panacea.NewTxBuilder(*e.reactor.QueryClient())
 
-	txBytes, err := generateTxBytes(msgVoteDataVerification, d.reactor.OracleAcc().GetPrivKey(), d.reactor.Config(), txBuilder)
+	txBytes, err := generateTxBytes(msgVoteDataVerification, e.reactor.OracleAcc().GetPrivKey(), e.reactor.Config(), txBuilder)
```
gyuguen: If merged my [PR](https://github.com/medibloc/panacea-doracle/pull/91/files#diff-1800206db049e385790536e9cc7a7c5eb8bc74dfaef4039c144c45dc5d0418abR25), you can used this function `txBuilder.GenerateTxBytes(...)`

event/datadeal/data_verification_event.go
```diff
@@ -53,61 +45,61 @@ func (d DataVerificationEvent) EventHandler(event ctypes.ResultEvent) error {
...
 	return nil
 }
 
-func (d DataVerificationEvent) decryptData(decryptedSharedKey, nonce, encryptedDataBz []byte) ([]byte, error) {
+func (e DataVerificationEvent) decryptData(decryptedSharedKey, nonce, encryptedDataBz []byte) ([]byte, error) {
 	decryptedData, err := crypto.DecryptWithAES256(decryptedSharedKey, nonce, encryptedDataBz)
 	if err != nil {
 		return nil, err
 	}
 	return decryptedData, nil
```
gyuguen: How about changing as direct return?
```suggestion
	return crypto.DecryptWithAES256(decryptedSharedKey, nonce, encryptedDataBz)
```

gyuguen (commented):
It looks good to merged function.
I left some comment :)

event/datadeal/data_verification_event.go
audtlr24: Good! Reflected in [dc45af6](https://github.com/medibloc/panacea-doracle/pull/95/commits/dc45af634288b9de9e1ce6b42d75b77bbeaf3625). Can we refactor our `broadcast` function in this same way? wdyt?

event/datadeal/data_verification_event.go
audtlr24: Thanks, I reflected this in [dc45af6](https://github.com/medibloc/panacea-doracle/pull/95/commits/dc45af634288b9de9e1ce6b42d75b77bbeaf3625). 👍

event/datadeal/data_verification_event.go
gyuguen: Good, But...
This function using grpcClient and simply priting result log.
So, the 'broadcast' function seems to vary depending on the each event.

audtlr24: I thought that different error messages for each event can be created outside the `broadcast` function, and `broadcast` just check `TxResponse.Code` and return height & hash if the transaction succeed. How about this?

panacea/tx.go
```diff
@@ -20,6 +21,17 @@ func NewTxBuilder(client QueryClient) *TxBuilder {
 	}
 }
 
+// GenerateTxBytes generates transaction byte array.
+func (tb TxBuilder) GenerateTxBytes(privKey cryptotypes.PrivKey, conf *config.Config, msg ...sdk.Msg) ([]byte, error) {
+	defaultFeeAmount, _ := sdk.ParseCoinsNormalized(conf.Panacea.DefaultFeeAmount)
```
0xHansLee: I couldn't remember why I skipped to handle error here.
Why don't you handle error here as well?

0xHansLee (approved):
Good. Readability seems better to me.
In case of the event of `SellData` and `RegisterOracle`, are there any issues related double subscription?

panacea/tx.go
audtlr24: I don't know why skip to handle error either. @gyuguen Is there any reason?

0xHansLee: I might write the code.. 😂. I don't remember..

gyuguen: I simply copied this code.
But it seems to wrong code. 
We need to handling error. 😂

gyuguen: So, I just fixed it in my PR . (https://github.com/medibloc/panacea-doracle/pull/91/commits/1467b54b33191b8e50ac6d880482a97874510337)

event/datadeal/data_verification_event.go
gyuguen: Therefore, Where would you like to create `broadcast` function?

panacea/tx.go
audtlr24: Okay, I fixed it, too. Thanks!

0xHansLee: Thank you guys!

event/datadeal/data_verification_event.go
gyuguen: And when a transaction fails, each event can have different error handling.
If only returned height and hash, we cannot handling error

0xHansLee: For now, we don't handle any error when tx fails, so I think it could be good.
If we make `broadcast` more general, I would make a function of `service` struct :)

gyuguen: OK.
If so, please refactorying this PR according to your opinion. @audtlr24

audtlr24: OK! After the PR #91 is applied, I will apply it at once :)

gyuguen (approved):
LGTM

service/service.go
```diff
@@ -127,3 +127,16 @@ func (s *Service) QueryClient() *panacea.QueryClient {
 func (s *Service) Ipfs() *ipfs.Ipfs {
 	return s.ipfs
 }
+
+func (s *Service) BroadcastTx(grpcClient *panacea.GrpcClient, txBytes []byte) (int64, string, error) {
+	resp, err := grpcClient.BroadcastTx(txBytes)
+	if err != nil {
+		return 0, "", fmt.Errorf("broadcast transaction failed. txBytes(%v)", txBytes)
+	}
```
0xHansLee: You don't need to pass the `grpcClient` since `Service` already has grpc client.

```suggestion
func (s *Service) BroadcastTx(txBytes []byte) (int64, string, error) {
	resp, err := s.GRPCClient().BroadcastTx(txBytes)
	if err != nil {
		return 0, "", fmt.Errorf("broadcast transaction failed. txBytes(%v)", txBytes)
	}
```

panacea/tx.go
```diff
@@ -20,6 +21,20 @@ func NewTxBuilder(client QueryClient) *TxBuilder {
 	}
 }
 
+// GenerateTxBytes generates transaction byte array.
+func (tb TxBuilder) GenerateTxBytes(privKey cryptotypes.PrivKey, conf *config.Config, msg ...sdk.Msg) ([]byte, error) {
```
inchori: I have a question. Why we use the `msg` data type `...sdk.Msg` instead of `sdk.Msg`?

inchori (approved):
lgtm!
I left an one comment.

service/service.go
audtlr24: Thanks! I fixed it.

panacea/tx.go
audtlr24: `TxBuilder` in `cosmos-sdk/client` interface define `SetMsgs(msgs ...sdk.Msg)` that can be set multi Msgs and multi signatures.  I think we use this interface to generate signed transaction, so we just follow this form.

inchori: Ah, I got it. Thanks for explain!

0xHansLee: In general, several `sdk.Msg`s(actions) can be included in a single transaction.
So the `TxBuilder` get multiple `sdk.Msg`s.

PR #96, opened 2022-11-03
feat: added voterUniqueID in oracleRegistrationVote transaction

gyuguen:
## Implement
When executing oracleRegistrationVote, `voterUniqueID` is added to parameter.
So I applied it.


event/oracle/event.go
```diff
@@ -29,6 +29,7 @@ func makeMsgVoteOracleRegistration(uniqueID, voterAddr, votingTargetAddr string,
 
 		registrationVote := &oracletypes.OracleRegistrationVote{
 			UniqueId:               uniqueID,
+			VoterUniqueId:          voterUniqueID,
```
0xHansLee, line 32: This might be extension of the previous [discussion](https://github.com/medibloc/panacea-core/pull/488#pullrequestreview-1167844466).

Here, two uniqueIDs are used: `uniqueID` and `voterUniqueID`.
The `uniqueID` is a uniqueID from event subscription. May I ask what's the purpose of using this uniqueID in vote?
I think we can put the oracle's uniqueID(`voterUniqueID`) into `UniqueId` field in `OracleRegistrationVote` struct.

gyuguen: Thank you.

I just reflected our discussion into the code.

https://github.com/medibloc/panacea-doracle/pull/96/commits/06523f0d0178f1206a6bca68f57e458814ff6781

gyuguen: For reference, when execute UpgradeOracle transaction, we shoud using OracleRegistration's uniqueID

0xHansLee: I got why we need two unique IDs
We use the same vote for both oracle registration and upgrade. 
But in case of upgrade, the two different uniqueIDs are required: old version and new version.
I understand now. Thx!

0xHansLee (approved):
good

audtlr24 (approved):
lgtm!

inchori (approved):
LGTM

PR #97, opened 2022-11-04
feat: added lightClient logLevel in config.toml

gyuguen:
## Implement
Current, we are using `info` logType in lightClient.
But this log is confusing because it looks like an error.
```
I[2022-11-04|01:19:15.241] error from light block request from primary, replacing... error="height requested is too high" height=527 primary=http{tcp://15.165.191.68:26657}
I[2022-11-04|01:19:16.385] error from light block request from primary, replacing... error="height requested is too high" height=527 primary=http{tcp://15.165.191.68:26657}
I[2022-11-04|01:19:17.533] error from light block request from primary, replacing... error="height requested is too high" height=527 primary=http{tcp://15.165.191.68:26657}
I[2022-11-04|01:19:18.677] error from light block request from primary, replacing... error="height requested is too high" height=527 primary=http{tcp://15.165.191.68:26657}
```

So I improve that it possible set lightClient logType in `config.toml`
If you set as light-client-log-type="error", will not print the light-client log until an error occurs.


audtlr24 (approved):
LGTM!

0xHansLee (approved):
Unfortunately, info log of light client is too noisy...
We have no choice. thank you.

inchori (approved):
LGTM

PR #101, opened 2022-11-09
feat: fix account sequence mismatch

### LINKED ISSUE #98: Error occured that "account sequence mismatch"
## Background
I had run two oracles, one is old oracle and the other oracle is the new oracle.
When an event occurs, both oracles subscribe at the same time.
And at the same time execute the transaction.
Two transactions are received in blockChain at the same time. 
Resultly, One transaction is success, another transaction is failed.
The failed reason is `account sequence mismatched`.

## Implement
When oracle starts, it gets the active uniqueID from the blockchain and compares one's own uniqueID.
* If two uniqueID match, Oracle starts subscribing to the vote event.
* Is not matched, Oracle doesn't subscribe to the vote event.
* In both case, It does subscribed OracleUpgradeEvent.

And when Oracle receives OracleUpgradeEvent, it compares one's own UniqueID.
* If two uniqueID match, Oracle starts subscribing to the vote event.
* Is not matched, Oracle stopping subscribe to the vote event.

gyuguen:
## Implement
I imporve like below:
* Added `OracleUpgradedEvent`
  * This event is raised when the OracleUpgrade is compelete.
  * If is not same one's own uniqueID and active uniqeID, all voteEvents change enable status to `false`
  * if not, all voteEvent change eanble status to `true`
* VoteEvent enable state is `true`, then the vote is execute. Otherwise, the voting will not be executed.
* All events added `prepare` step before subscribing.


cmd/doracled/cmd/start.go
```diff
@@ -29,11 +30,16 @@ func startCmd() *cobra.Command {
...
 			defer svc.Close()
 
-			err = svc.StartSubscriptions(
+			voteEvents := []event.Event{
 				oracleevent.NewRegisterOracleEvent(svc),
 				oracleevent.NewUpgradeOracleEvent(svc),
 				datadealevent.NewDataVerificationEvent(svc),
 				datadealevent.NewDataDeliveryVoteEvent(svc),
+			}
+			upgradedOracleEvent := oracleevent.NewUpgradedOracleEvent(svc, voteEvents)
```
gyuguen, line 39: `UpragedOracleEvent` need to voteEvents parameter.

event/oracle/upgraded_oracle_event.go
```diff
@@ -0,0 +1,70 @@
...
+
+func NewUpgradedOracleEvent(r event.Reactor, voteEvents []event.Event) *UpgradedOracleEvent {
+	return &UpgradedOracleEvent{
+		reactor:    r,
+		enable:     true,
+		voteEvents: voteEvents,
+	}
+}
+
+func (e *UpgradedOracleEvent) Prepare() error {
```
gyuguen, line 32: This function is called before each event subscribes.

PR #102, opened 2022-11-10
test: ipfs unit test

inchori:
Implement a IPFS test code with docker.
- Add a test code for `Add` and `Get` method.


ipfs/ipfs_test.go
```diff
@@ -1,46 +1,63 @@
...
+
+type testdata struct {
+	Name        string `json:"name"`
+	Description string `json:"description"`
+}
+
+func (suite *ipfsTestSuite) TestIpfsAdd() {
+	newIpfs := ipfs.NewIpfs(suite.IpfsEndPoint(5001))
+
+	testData := &testdata{
```
gyuguen, line 32: I think it would be good to change the value in testData.
Because if you add the same data, ipfs will always response the same cid.

ipfs/suite/suite.go
```diff
@@ -0,0 +1,77 @@
+package suite
```
gyuguen: How about moving this file to `integration/suite/suite_ipfs.go`?

gyuguen (commented):
It's very nice!
I left a few comment

gyuguen (approved):
and LGTM

ipfs/ipfs_test.go
inchori: Does it mean use a data such as random number thing where I added in [9ba51cf](https://github.com/medibloc/panacea-doracle/pull/102/commits/9ba51cf5b3d9923b0be4de4695e4e9e371ebe5d0)?

gyuguen: Yes!

